PR #17, opened 2013-07-02
Tentative fix for #77 and #154

Zlika:
Tentative fix for #77 and #154. The default behaviour ("Accept-Encoding: gzip,deflate") is kept to avoid breaking backward compatibility.
The only problem I see with this fix is that it uses an internal type (com.jayway.restassured.internal.http.ContentEncoding) in the public API of RequestSpecification, but this problem already exists for ObjectMapperType. Maybe this type could be moved outside the internal package ? (I would like to avoid creating a new public "ContentEncoding" that would be a mere copy-paste of the original one)


karthik008:
Hi I am still getting error stating that "java.util.zip.ZipException: incorrect header check",Iam using Rest assured 3.0

Response r = io.restassured.RestAssured.given().headers("content-type","application/json").headers("Accept","application/json").contentType(ContentType.JSON).body(updateBody).when().post("http://Common.domain.com/api/api/TestItem/678566/3.3.90.57/test/").then().extract().response();

PR #39, opened 2015-05-10
[#307] Added async support for Spring MockMvc

marcingrzejszczak:
Initial proposal of an API to add async support for Spring MVC

example of a spec with async support:

```
given().
                asyncTimeout(10).
                body("a string").
        when().
                post("/stringBody").
        then().
                body(equalTo("a string"));
```

Where `asyncTimeout` API looks as follows (example for Specification):

```
/**
     * Specify the timeout during which you will wait for async response.
     *
     * @param timeoutInMillis timeout in millis to wait for async response.
     * @return A {@link MockMvcAuthenticationSpecification}.
     */
    MockMvcRequestSpecification asyncTimeout(long timeoutInMillis);
```

Under the hood what it does is what is described in Spring's API - org.springframework.test.web.servlet.request.MockMvcRequestBuilders#asyncDispatch

```
/**
	 * Create a {@link RequestBuilder} for an async dispatch from the
	 * {@link MvcResult} of the request that started async processing.
	 * <p>Usage involves performing one request first that starts async processing:
	 * <pre class="code">
	 * MvcResult mvcResult = this.mockMvc.perform(get("/1"))
	 *	.andExpect(request().asyncStarted())
	 *	.andReturn();
	 *  </pre>
	 * <p>And then performing the async dispatch re-using the {@code MvcResult}:
	 * <pre class="code">
	 * this.mockMvc.perform(asyncDispatch(mvcResult))
	 * 	.andExpect(status().isOk())
	 * 	.andExpect(content().contentType(MediaType.APPLICATION_JSON))
	 * 	.andExpect(content().string("{\"name\":\"Joe\",\"someDouble\":0.0,\"someBoolean\":false}"));
	 * </pre>
	 * @param mvcResult the result from the request that started async processing
	 */
```

So the usage to retrieve the result from the async dispatch:

```
MvcResult mvcResult;
        if (mockMvcAsyncConfig != null && mockMvcAsyncConfig.isAsync()) {
            MvcResult startedAsyncRequestProcessing = perform.andExpect(request().asyncStarted()).andReturn();
            startedAsyncRequestProcessing.getAsyncResult(mockMvcAsyncConfig.getTimeoutInMs());
            mvcResult = mockMvc.perform(asyncDispatch(startedAsyncRequestProcessing)).andReturn();
        } else {
            mvcResult = perform.andReturn();
        }
        return mvcResult;
```

where `mockMvcAsyncConfig` contains the information from the API - `long timeoutInMillis`.

I've already tested that in a project where we would need this feature and it seems to be working.


johanhaleby:
Thanks a lot for the helping out, I really really appreciate it. Some questions and comments though:

Do I understand it correctly that by setting "asyncTimeout" the request turns into an asynchronous one? My question is, wouldn't it be easier to understand from a user perspective to have an async version of each request method? So instead of doing:

```java
given().
        asyncTimeout(10).
        body("a string").
when().
        post("/stringBody").
then().
        body(equalTo("a string"));
```

you would do:

```java
given().
        asyncTimeout(10).
        body("a string").
when().
        postAsync("/stringBody").
then().
        body(equalTo("a string"));
````
and `postAsync` is what make the request go async and not "asyncTimeout". You could also have the option of doing just:

```java
given().
        body("a string").
when().
        postAsync("/stringBody").
then().
        body(equalTo("a string"));
````
and then REST Assured would use a default timeout. What do you think about this?

**Comment by marcingrzejszczak** [comment]
Hi!

Yes - you understand it correctly. By adding the 'asyncTimeout' the request turns into asynchronous one.

I was thinking about adding `(methodName)Async` but I thought that it would mean some redundancy. With my approach you add timeout only once and then when performing a request the choice is made whether the request should be async or not.

As for a default timeout I was thinking about adding the `async()` method that would signify that there is an asynchronous request coming with a default timeout.

But of course you have a better understanding of user's needs ;) I can prepare a solution with all the methods with a Async suffix.

**Comment by marcingrzejszczak** [comment]
BTW in that approach what should happen if somebody does:

```
given().
        asyncTimeout(10).
        body("a string").
when().
        post("/stringBody").
then().
        body(equalTo("a string"));
```

THere is `asyncTimeout` in the setup and no `Async` in the `when` part?

**Comment by johanhaleby** [comment]
I think you're on to something regarding:

```java
given().
        asyncTimeout(10).
        body("a string").
when().
        post("/stringBody").
then().
        body(equalTo("a string"));
```

This shows that the API I suggested probably wouldn't be a good idea. 

But if you instead would do like this:

```java
given().
        body("a string").
when().
        postAsync("/stringBody", 10).
then().
        body(equalTo("a string"));
```

I.e. you pass the timeout as an argument to the `postAsync` method. One could argue that you would loose readability ("10" looks like a magic number until you read the parameter definitions or Javadoc for postAsync). But I think it's legitimate given that the timeout is something that you want to set often. 

But even so adding "Async" to each request method (get(), post() etc) will add A LOT of new methods since each of these methods already have a lot of overloaded methods. Another approach would be to do the following:

```java
given().
        body("a string").
when().
        async().post("/stringBody", 10).
then().
        body(equalTo("a string"));
```

(where 10 is optional). async() could return a new interface, perhaps called something like AsyncRequestSender, that contains all the request methods. It could probably work in a similar way that logging does. When you call `log()` it returns a new interface that allows you to specify what to log. For example "log().headers()" and then it returns the Request-/Response specification again.

Another thing regarding the `MockMvcAsyncConfig`. I think it ought to extend the `com.jayway.restassured.config.Config` interface and work the same way as the rest of the configuration classes in RA. This means that you can more easily reuse your async configuration in different scenarios (note that you should only add this config in the MockMvc project and thus to the RestAssuredMockMvcConfig). See https://code.google.com/p/rest-assured/wiki/Usage#Specification_Re-use if you're unfamiliar with this. If this seems overwhelming then don't worry, I'll try to help you out :)

What do you think?

johanhaleby:
Btw, wouldn't it be quite nice to make use of `java.util.concurrent.TimeUnit` when specifying timeout as a compliment to just the number of milliseconds? For example:

```java
given().
        body("a string").
when().
        async().post("/stringBody", timeout(2, TimeUnit.SECONDS)).
then().
        body(equalTo("a string"));
```

Where "timeout" could be statically imported from somewhere.. Perhaps the `MockMvcAsyncConfig` class. This means that the async version of post would take a MockMvcAsyncConfig as an argument as well. Note sure if this is a good idea though since it would make MockMvcAsyncConfig special in regard to the other configs which are specified like this:

```java
given().config(config().x(new XConfig())). ..
```

marcingrzejszczak:
Hi!

I didn't know about this `com.jayway.restassured.config.Config` so of course I'll adjust the `MockMvcAsyncConfig` to implement this.

When I proposed my change I chose a solution that was the easiest to implement :) I understand that the API wasn't necessarily the best so if I understand correctly you would suggest this:

```
given().
        body("a string").
when().
        async().post("/stringBody", 10).
then().
        body(equalTo("a string"));
```

SO let's agree that I'll try to reimplement the current approach to support this.

As for `timeout(2, TimeUnit.SECONDS)` - it's feasible but the visibility of configs is also troubling for me. However one could write

```
given().
        body("a string").
when().
        async().post("/stringBody", SECONDS.toMillis(10)).
then().
        body(equalTo("a string"));
```

and it looks relatively ok again :) 

WDYT?

marcingrzejszczak:
Ok I've done both suggestions.

1) `MockMvcAsyncConfig` implements `com.jayway.restassured.config.Config`
2) I've changed the API

Now it looks like this:

```
 @Test public void
    can_supply_string_as_body_for_async_post() {
        given().
                body("a string").
        when().
            async().
                post(withTimeout(10, MILLISECONDS), "/stringBody").
        then().
                body(equalTo("a string"));
    }
```

How does it look like internally?

`com.jayway.restassured.module.mockmvc.specification.MockMvcAsyncRequestSender` in almost all methods take a `Timeout` parameter. It couldn't be added as last because sometimes you need to pass a list of objects so the compiler would scream. Also if it's passed at the beginning by means of the static method `withTimeout` then it seems more descriptive.

So this `Timeout` is a nested class in `com.jayway.restassured.module.mockmvc.specification.MockMvcAsyncRequestSender`. That's why the autocompletion from IDE will automatically find the class and its static method `withTimeout`.

Also if you provide async() and don't provide any option then a default of 1 second will be taken.

I haven't added more tests because first I want to be sure that we're on the same page in terms of the API.

johanhaleby:
I think you're right that the timeout must be specified first to avoid screwing up the var arg parameters. I also like that you don't need to specify a timeout and if so a default timeout is used.

However I'm not sure if we need a special "Timeout" class. I would prefer that the request methods in `MockMvcAsyncRequestSender` takes an instance of `MockMvcAsyncConfig`. It could still read the same:

```java
@Test public void
can_supply_string_as_body_for_async_post() {
    given().
            body("a string").
    when().
        async().
            post(withTimeout(10, MILLISECONDS), "/stringBody").
    then().
            body(equalTo("a string"));
}
```

But `withTimeout` is a static method in `MockMvcAsyncConfig` that simply returns a new instance of the `MockMvcAsyncConfig` with the given timeout. What do you think about that?

Regarding the amount of code. One idea to reduce the amount of code would be to avoid letting the request methods in `MockMvcAsyncRequestSender` altogether and instead simply specify a `MockMvcAsyncConfig` in the same way that you do with all other configs. For example:

```java
@Test public void
can_supply_string_as_body_for_async_post() {
    given().             
            config(RestAssuredMockMvc.config().asyncConifg(MockMvcAsyncRequestSender.withTimeout(10, MILLISECONDS)).
            body("a string").
    when().
        async().
            post("/stringBody").
    then().
            body(equalTo("a string"));
}
```

Using static imports it would look slightly better:

```java
@Test public void
can_supply_string_as_body_for_async_post() {
    given().             
            config(config().asyncConifg(withTimeout(10, MILLISECONDS)).
            body("a string").
    when().
        async().
            post("/stringBody").
    then().
            body(equalTo("a string"));
}
```

With this approach my hope is that we can reuse the `RequestSender` interface (i.e. `async()` returns an instance of  `RequestSender`) and thus there's no use for the `MockMvcAsyncRequestSender` interface. 

` given().config(config().asyncConifg(withTimeout(10, MILLISECONDS)). ..` is arguably not as nice as the previous example of `post(withTimeout(10, MILLISECONDS), "/stringBody"). ..` but it's more consistent with how all other configs work in REST Assured. But it would be possible to make it nice as well. If we take the example of logging again, you can either specify logging using the `com.jayway.restassured.config.LogConfig` (using `given().config(config().logConfig(new LogConfig(ALL)). ..`) or using the DSL/API (`given().log().all()`). We could do the same thing for the async use case. For example something like this:

```java
@Test public void
can_supply_string_as_body_for_async_post() {
    given().             
            body("a string").
    when().
        async().
            withTimeout(10, MILLISECONDS).
            post("/stringBody").
    then().
            body(equalTo("a string"));
}
```

Implementation-wise my hope is that `async()` returns an interface that can extend from `RequestSender` but also allow you to configure a `MockMvcAsyncConfig` under the hood. When the user calls `post("/stringBody")` an instance of `MockMvcAsyncConfig` is created with the configured values and added to the `RequestSpecification` *before* the async call is actually executed. 

Actually to make even more consistent with how other parts of the API works I think the DSL should actually read like this:

```java
@Test public void
can_supply_string_as_body_for_async_post() {
    given().             
            body("a string").
    when().
        async().
            with().timeout(10, MILLISECONDS).
            post("/stringBody").
    then().
            body(equalTo("a string"));
}
```

`with` is just a syntactic sugar method that returns `this`. This way users can choose for whether they like to go with a more english like DSL (using `with`) or just skip it altogether. If `MockMvcAsyncConfig` contains more stuff than just timeout in the future the DSL could "scale" to look something like this:

```java
@Test public void
can_supply_string_as_body_for_async_post() {
    given().             
            body("a string").
    when().
        async().
            with().timeout(10, MILLISECONDS).and().x("something").and().y(2).
            post("/stringBody").
    then().
            body(equalTo("a string"));
}
```

Which would be the same as just writing:

```java
@Test public void
can_supply_string_as_body_for_async_post() {
    given().             
            body("a string").
    when().
        async().
            timeout(10, MILLISECONDS).x("something").y(2).
            post("/stringBody").
    then().
            body(equalTo("a string"));
}
```

I.e. `and()` is also just syntactic sugar. What do you think about this?

Sorry if this is causing you a lot of rewrites. I can try to help out with the implementation as well if you like.

Btw, I've found a bug in the `MockMvcAsyncConfig`, the `isUserConfigured` method always returns `false`.

marcingrzejszczak:
Ok so API looks as follows:

Configuration in the `given` section

```java
 @Test public void
    can_supply_string_as_body_for_async_post_with_config_in_given() {
        given().
                config(newConfig().asyncConfig(withTimeout(10, TimeUnit.MILLISECONDS))).
                body("a string").
        when().
                post("/stringBody").
        then().
                body(equalTo("a string"));
    }
```

Async in the `when` part:

```java
    @Test public void
    can_supply_string_as_body_for_async_post() {
        given().
                body("a string").
        when().
            async().
                post("/stringBody").
        then().
                body(equalTo("a string"));
    }
```

Async with config in the `when` part

```java
given().
                    body("a string").
            when().
                async().
                    with().timeout(100, TimeUnit.MILLISECONDS).
                    post("/tooLongAwaiting").
            then().
                    body(equalTo("a string"));
```

The thing that I'd prefer to do otherwise is that when we do the `async().with()` we could go to another interface that gives you only configuration options (like this `timeout` stuff) and could end with some `andThen()` method that would allow you to come back to the `RequestSender` extension. Sth like:

```java
given().
                    body("a string").
            when().
                async().
                    with().timeout(100, TimeUnit.MILLISECONDS).andThen().
                    post("/tooLongAwaiting").
            then().
                    body(equalTo("a string"));
```

But without that the API looks as you wanted.

Ans of course I fixed the bug with the Configuration.

johanhaleby:
Wow that was fast!

I need to think about "andThen" for a little bit. It may actually be less confusing for the user. Otherwise you could do something like ".. async().with().post(..)" which is not intended.

Great work :+1:

marcingrzejszczak:
Thanks :) 

Exactly that was my idea behind this concept. So that people don't get confused.

That's actually pretty simple to do so it's your call.

johanhaleby:
I'm thinking about this minor change:

```java
given().
        body("a string").
when().
    async().
        with().timeout(100, TimeUnit.MILLISECONDS).and().then().
        post("/tooLongAwaiting").
then().
        body(equalTo("a string"));
```

Where `and` is optional (syntactic sugar). I think that would be more consistent with the rest of the API. Do you think that would be a good idea or can you spot any concerns?

marcingrzejszczak:
I think that it's a very good change :) I've just pushed the change.

Now:

```
@Test public void
    can_supply_string_as_body_for_async_post_with_config_in_given() {
        given().
                config(newConfig().asyncConfig(withTimeout(10, TimeUnit.MILLISECONDS))).
                body("a string").
        when().
                post("/stringBody").
        then().
                body(equalTo("a string"));
    }

    @Test public void
    exception_will_be_thrown_if_async_data_has_not_been_provided_in_defined_time_with_config_in_given() {
        // given
        Exception exception = null;

        // when
        try {
            given().
                    config(newConfig().asyncConfig(withTimeout(0, TimeUnit.MILLISECONDS))).
                    body("a string").
            when().
                    post("/tooLongAwaiting").
            then().
                    body(equalTo("a string"));
        } catch (IllegalStateException e) {
            exception = e;
        }

        // then
        assertThat(exception).isNotNull().hasMessageContaining("was not set during the specified timeToWait=0");
    }

    @Test public void
    can_supply_string_as_body_for_async_post() {
        given().
                body("a string").
        when().
            async().and().then().
                post("/stringBody").
        then().
                body(equalTo("a string"));
    }

    @Test public void
    exception_will_be_thrown_if_async_data_has_not_been_provided_in_defined_time() {
        // given
        Exception exception = null;

        // when
        try {
            given().
                    body("a string").
            when().
                async().
                    with().timeout(0, TimeUnit.MILLISECONDS).and().then().
                    post("/tooLongAwaiting").
            then().
                    body(equalTo("a string"));
        } catch (IllegalStateException e) {
            exception = e;
        }

        // then
        assertThat(exception).isNotNull().hasMessageContaining("was not set during the specified timeToWait=0");
    }
```

johanhaleby:
Nice! I just wonder, is it required to use `and().then()` when you don't want to specify a config:

```java
.. .when().async().and().then().post("/x"). ..
``` 

or can you also do just:

```java
.. .when().async().post("/x"). ..
``` 

if you don't need to specify a configuration? I think that latter would be much nicer.

marcingrzejszczak:
We could do it as you suggest but we'd have to have a word that would
initialize a config.

Sth like

```when().async().withConfig()```

That way we would go through another interface.

WDYT?

**Comment by marcingrzejszczak** [comment]
Ok you were right - I fixed it as you wanted and now it looks nice. You can do it like this:

```java

 @Test public void
    can_supply_string_as_body_for_async_post() {
        given().
                body("a string").
        when().
            async().and().then().
                post("/stringBody").
        then().
                body(equalTo("a string"));
    }
```

but also:

```java
    @Test public void
    can_supply_string_as_body_for_async_post_without_syntactic_sugar() {
        given().
                body("a string").
        when().
            async().
                post("/stringBody").
        then().
                body(equalTo("a string"));
    }
```

Cool thing is that it's done in such a way that once you do a `post` you can't go back to confguring so people will not create odd dsl calls.

johanhaleby:
Great work!!! Thanks A LOT for all your help.

johanhaleby:
There are some things that I'd like to change internally before merging the pull request. I'll try to find time to this tomorrow.

marcingrzejszczak:
Awesome - great to be a contributor to such a cool project :) 

We need it at our project here - https://github.com/Codearte/accurest so it will be a benefit for both sides.

johanhaleby:
I've merged it to master now! I've refactored some internals but also changed the API a bit:

1. I've added a `timeout` method that takes a duration AND a timeunit in the async api.
2. I've made `async` mandatory for async requests. This means that only specifying an `AsyncConfig`  (I renamed it from `MockMvcAsyncConfig`) will have no effect unless you also do `.. when().async().`.
3. I squashed all your commits into one

It's integrated to the master branch now. I've also deployed a new snapshot. If you'd like to try it out please depend on version `2.4.2-SNAPSHOT` after having added the following repo (maven):

```xml
<repositories>
        <repository>
            <id>sonatype</id>
            <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
            <snapshots />
        </repository>
</repositories>
```

Again, thanks a lot for all your help and great suggestions :+1:. I'm glad you find REST Assured useful.  I'll be happy to work with you again. 

Btw, I googled your name and I found you're into microservices (who isn't these days? :)). I have an idea to create a module to REST Assured that generates <a href="https://github.com/DiUS/pact-jvm">pact</a> provider contracts when running the tests. Perhaps this would interest you as well (not saying you need to implement it)? Not sure if it's feasible though.

marcingrzejszczak:
Hi!

I haven't noticed your message :D

We are using this in our project Accurest - https://github.com/Codearte/accurest which is a solution much simpler than packt. Maybe take a look at how we approach things and maybe it would be more interesting for you :)

Accurest actually already uses Rest assured to generate tests of a contract provided in a DSL. If you have any questions about how it's working just ping me :)

johanhaleby:
Thanks for the hint! I'll take a loot at it.

marcmyers:
I've been trying to get this to work in my project, I think I'm importing incorrectly because it can't resolve async(). My project is mavenized, I've been trying to do a PUT and then do a GET while the PUT is still waiting for a response. Are there any mavenized projects using this that would serve as a good example?

johanhaleby:
Are you sure you're using the latest version and that you also have the rest assured spring-mock-mvc module in the classpath? If you do, are you really importing the static methods from `RestAssuredMockMvc` and not `RestAssured`? Have a look at the documentation for static imports.

PR #40, opened 2015-06-25
fix: quotations in charset parameter

rbraeunlich:
-charsets that are quoted will now be recognized by removing the quotes, so that e.g. charset="UTF-8" works

Solves Issue 413


johanhaleby:
Thanks for the pull request

PR #593, opened 2015-09-27
fix #592 GPath namespace handling

### LINKED ISSUE #592: GPath inconsistency in XmlPath with namespaces
Without XML namespaces the GPath needs to include the root object name but when using namespaces it must not.

Assuming the following XML without namespaces:

```xml
<licensesResponse>
  <license-active>true</license-active>
</licensesResponse>
```

the corresponding GPath needs to be written as

```java
...
  .then()
     .body("'licensesResponse.license-active'", equalTo("true"))
```

but if the xml is namespaced

```xml
<licensesResponse xmlns='http://example.com/licenses'>
   <license-active>true</license-active>
</licensesResponse>
```

the corresponding GPath needs to be written as

```java
        given().config(config()
            .xmlConfig(xmlConfig()
                .namespaceAware(true)
                .declareNamespace("licenses", "http://example.com/licenses"))
        )
...
        .then()
               .body("licenses:license-active", equalTo("true"))
        ;
```

this seems due to the call to 

```java
return (T) xmlAssertion.getResult(input, convertToJavaObject, !getXmlPathConfig().hasDeclaredNamespaces());
```

in method `com.jayway.restassured.path.xml.XmlPath#getFromPath`

I am wondering what the rational behind `rootEvaluation = !getXmlPathConfig().hasDeclaredNamespaces()` is?

eeichinger:
fixes #592 by always resolving an gpath from the root. Also updates various related unit tests that used to "accidentially" pass incorrectly

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -50,6 +50,20 @@ Change log next version
 * It's now possible to specify default mime subtype for multipart content-type. Use the MultiPartConfig#defaultSubtype(..) method.
   Default is "form-data" which results in a content-type of "multipart/form-data". This also works for the MockMvc module.
 * Upgraded JUnit to 4.12
+* fix evaluation of GPath expression from root if xml namespaces are declared (issue 592). For example given the following XML:
+      <foo xmlns:ns="http://localhost/">
+         <bar>sudo </bar>
+         <ns:bar>make me a sandwich!</ns:bar>
+      </foo>
+  you can verify it like this:
+      given().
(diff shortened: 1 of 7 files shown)


johanhaleby:
Thanks a ton!!

eeichinger:
@johanhaleby need to update the wiki as well in the section about xml namespaces - unfort. it doesn't seem I can create a pull request for wiki changes

johanhaleby:
I had planned do that when the next version is released. Thanks for reminding me.

Issue #544, opened 2015-07-21
Cannot set a request body for a GET method


johanhaleby:
_From [eskato...@gmail.com](https://code.google.com/u/115671622875546989676/) on March 05, 2015 17:22:52_

Using rest-assured 2.4.0, one can't send GET requests with a body.

It should work but it throw saying "Cannot set a request body for a GET method".
Here an excerpt of the stacktrace:

    java.lang.IllegalArgumentException: Cannot set a request body for a GET method
        at com.jayway.restassured.internal.http.HTTPBuilder$RequestConfigDelegate.setBody(HTTPBuilder.java:1076)
        at com.jayway.restassured.internal.http.HTTPBuilder$RequestConfigDelegate.setBody(HTTPBuilder.java:1061)

HTTP do not forbid bodies on GET requests.
Some largely used products like ElasticSearch require GET requests with bodies.

_Original issue: http://code.google.com/p/rest-assured/issues/detail?id=389_

johanhaleby:
_From [johan.ha...@gmail.com](https://code.google.com/u/105676376875942041029/) on March 16, 2015 03:53:49_

Then I think they're misusing the HTTP GET verb. You CAN send a body in a GET request but shouldn't. See for example http://stackoverflow.com/questions/978061/http-get-with-request-body . Even Roy Fielding (the inventor of REST) says this:

"So, yes, you can send a body with GET, and no, it is never useful to do so."

**Status:** WontFix

johanhaleby:
_From [eskato...@gmail.com](https://code.google.com/u/115671622875546989676/) on March 16, 2015 06:53:08_

Sure but as it is not forbidden, some products do it.

I think that `rest-assured`, as a testing oriented library, should allow us to do so.

What is forbidden is for a server to parse it and change it's behaviour based on its content.

Please reconsider this.

johanhaleby:
_From [johan.ha...@gmail.com](https://code.google.com/u/105676376875942041029/) on March 16, 2015 08:54:04_

I see your point and I'm not totally against this. However I don't think Apache HTTP Client supports this by default either (and RA builds on top of AHC) so this means that it'll probably be a bit hard to implement (other things that RA supports that AHC doesn't such as non-value parameters were quite hard to implement). But if you want to help out you could perhaps find out how to do this with Apache HTTP Client then it would be much easier for me to implement in RA.

**Status:** Accepted  
  **Labels:** -Type-Defect Type-Enhancement

PR #603, opened 2015-10-16
Fix for #544

quitschibo:
I added support for a GET request body. I also had to update the com.squareup.okhttp.mockwebserver dependency; in the older version the mockserver checks, if the GET request has a body and then it throws an exception. In the new version this behavior has changed.


johanhaleby:
You sir are awesome! Thanks a lot for this pull request, it's something I've wanted to support for a long time.

PR #604, opened 2015-10-16
Fix for #566

quitschibo:
Response content-type validation now works correctly, even if the response body is empty.


johanhaleby:
Thanks again!!

PR #608, opened 2015-10-28
Added getUUID to JsonPath and XmlPath

librucha:
In many cases UUID is returned as identificator of resource and is unlesses convert it from String.


johanhaleby:
Thanks for the pull request

PR #614, opened 2015-11-05
Modify HttpClientConfig so parameter methods respect previous configuration. Fixes #612

### LINKED ISSUE #612: HttpClientConfig setParams returns new HttpClientConfig object which does not respect other config settings
When configuring HttpClientConfig, methods such as httpClientFactory return a new HttpClientConfig object constructed using all current configuration options (params, multipart mode, instance reuse). However, setParam, addParams, and withParams return a new HttpClientConfig with only the new parameters set, other configurations are not persisted.

This means that if you want to configure multiple aspects of a HttpClientConfig you need to make sure setParams is the first call, otherwise it will wipe out all previous ones:

```
RestAssured.config = RestAssured.config()
                .httpClient(
                        HttpClientConfig.httpClientConfig()
                                .setParam("http.protocol.cookie-policy", "compatibility")
                                .httpClientFactory(new HttpClientConfig.HttpClientFactory() {
                                    @Override
                                    public HttpClient createHttpClient() {
                                        return httpClient;
                                    }
                                })
                );
```

It would make sense for this to work irrespective of order:

```
RestAssured.config = RestAssured.config()
                .httpClient(
                        HttpClientConfig.httpClientConfig()
                                .httpClientFactory(new HttpClientConfig.HttpClientFactory() {
                                    @Override
                                    public HttpClient createHttpClient() {
                                        return httpClient;
                                    }
                                })
                               .setParam("http.protocol.cookie-policy", "compatibility")
                );
```

However in this case, the configuration of a custom httpClientFactory is lost.

ajclarkson:
Alterations to HttpClientConfig to address issue referenced in issue [https://github.com/jayway/rest-assured/issues/612] 

I added a few tests to make sure this was behaving as expected now and all is fine. Also addressed the bug with the addParams method only setting the new parameters, not merging them with existing ones

### DIFF
--- a/rest-assured/pom.xml
+++ b/rest-assured/pom.xml
@@ -121,7 +121,13 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
+            <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>1.7.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


johanhaleby:
Thanks a lot for the pull request. Great work.

ajclarkson:
No worries, thanks for the library!

Issue #633, opened 2016-01-20
closeIdleConnectionsAfterEachResponse closes the response before consuming is possible


mdzhigarov:
Im using the following configuration to reproduce the issue: 

RestAssured.config = RestAssured.config().connectionConfig(new ConnectionConfig().closeIdleConnectionsAfterEachResponse());
RequestSpecification request = given().relaxedHTTPSValidation()
                         .contentType(Constants.XML_CONTENT_TYPE)
                         .auth().basic("usr",
                                       "pass");
request .contentType("application/json");
request.accept("application/json");
request.body(...);
request.when().post(restURL).then().extract().response().asString();

The asString() call throws exception:
java.net.SocketException: Socket is closed
	at sun.security.ssl.SSLSocketImpl.checkEOF(SSLSocketImpl.java:1520)
	at sun.security.ssl.AppInputStream.read(AppInputStream.java:95)
	at org.apache.http.impl.io.AbstractSessionInputBuffer.fillBuffer(AbstractSessionInputBuffer.java:160)
	at org.apache.http.impl.io.SocketInputBuffer.fillBuffer(SocketInputBuffer.java:84)
	at org.apache.http.impl.io.AbstractSessionInputBuffer.readLine(AbstractSessionInputBuffer.java:273)
	at org.apache.http.impl.conn.LoggingSessionInputBuffer.readLine(LoggingSessionInputBuffer.java:116)
	at org.apache.http.impl.io.ChunkedInputStream.getChunkSize(ChunkedInputStream.java:240)
	at org.apache.http.impl.io.ChunkedInputStream.nextChunk(ChunkedInputStream.java:206)
	at org.apache.http.impl.io.ChunkedInputStream.read(ChunkedInputStream.java:169)
	at java.util.zip.InflaterInputStream.fill(InflaterInputStream.java:238)
	at java.util.zip.InflaterInputStream.read(InflaterInputStream.java:158)
	at java.util.zip.GZIPInputStream.read(GZIPInputStream.java:117)
	at org.apache.http.conn.EofSensorInputStream.read(EofSensorInputStream.java:137)
	at org.apache.http.conn.EofSensorInputStream$read.call(Unknown Source)
	at com.jayway.restassured.internal.RestAssuredResponseOptionsGroovyImpl.convertStreamToByteArray(RestAssuredResponseOptionsGroovyImpl.groovy:430)
	at sun.reflect.GeneratedMethodAccessor158.invoke(Unknown Source)
[...]

PR #639, opened 2016-02-02
Replace deprecatet HttpEntity.consume with EntityUtils.consume

oxygen0211:
fixes issue #633

### DIFF
--- a/rest-assured/pom.xml
+++ b/rest-assured/pom.xml
@@ -64,6 +64,12 @@
             <version>4.5.1</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpcore</artifactId>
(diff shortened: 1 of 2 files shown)


oxygen0211:
Apparently, in certain cases (at least in the integration tests), EntityUtils is not available. Have to check this.

johanhaleby:
Thanks for the pull request. In what scenarios are `EntityUtils` not available? Do you mean in some version of Apache HTTP Client?

oxygen0211:
Turned out it was an error on my side, I missed out an import that wasn't added automatically by Eclipse (I'm not doing much groovy and thus my env isn't really prepared for it). With ec728f6, those concerns are gone

johanhaleby:
Oh OK! Great I'll bring it on.

PR #659, opened 2016-03-20
Upgrading scribe dependencies to latest version 2.4.0

aniket-21:


johanhaleby:
Thanks for pull request. Will this affect backward compatibility in some way?

aniket-21:
I don't think so

johanhaleby:
:+1:

PR #660, opened 2016-03-21
Config Flag to decide whether empty token should be added to Base String

aniket-21:


rest-assured/src/main/java/com/jayway/restassured/internal/http/AuthConfig.java
```diff
@@ -196,23 +196,25 @@ public void oauth2(String accessToken, OAuthSignature signature) {
...
             this.oauthConfig = new OAuthConfig(consumerKey, consumerSecret,
                     null, getOAuthSigntureType(signature), null, null, null, null, null, null, null);
             this.token = new OAuth1AccessToken(accessToken, secretToken);
             this.signature = signature;
+            this.addEmptyTokenToBaseString = addEmptyTokenToBaseString;
         }
 
         public OAuthSigner(String accessToken, OAuthSignature signature) {
-            this.token = new OAuth1AccessToken(accessToken, "");
+            this.token = new OAuth2AccessToken(accessToken, "");
```
johanhaleby: Is this correct? Changing from OAuth1 to OAuth2!?

aniket-21: Yes it is. Before I upgraded to latest scribe the class was "**_Token_**" so it did not matter if it was Oauth1 or Oauth2 Signer. But when I updated, I updated both with Oauth1AccessToken and later realized that **_OAuthSigner(String accessToken, OAuthSignature signature)_** is being specifically being used for Oauth2

johanhaleby:
Cool! Thanks a lot for your persistent support @aniket-21. Really helpful. What should I write in the [changelog](https://raw.githubusercontent.com/jayway/rest-assured/master/changelog.txt)? Does something like this look ok?

> Added ability to specify an empty access token for OAuth1 authentication. Use com.jayway.restassured.config.OAuthConfig#addEmptyTokenToBaseString(boolean)

And one last thing, shouldn't the method on OAuthConfig be called `addEmptyAuthTokenToBaseString` to make it more explicit?

aniket-21:
Agree, we should rename the method in OAuthConfig. Also the changelog description looks good.
Thanks to you to @johanhaleby for your guidance. Cheers!

johanhaleby:
I'll fix it! Thanks again.

PR #662, opened 2016-03-22
Disambiguated full-dist and module jar names for rest-assured jars, a…

### LINKED ISSUE #661: Disambiguate full/combined/all-modules dist jar and rest-assured module jar
The dist build currently creates a couple of non-Mavenized artifacts, a deps zip and a dist zip, and an uber-jar style all-modules rest-assured jar.  The latter is named e.g. `rest-assured-2.9.1-SNAPSHOT.jar` and contains all modules `rest-assured`, `rest-assured-common`, `json-path` and `xml-path`.

The only issue with that is that the aforementioned dist uber/all jar and the `rest-assured` module jar have identical filenames i.e. `rest-assured-2.9.1-SNAPSHOT.jar`.  If you have been handling/testing the all-modules jar, then switch to Maven, this might be somewhat confusing - the jars have the same name, but the Maven/module jar (`com.jayway.restassured:rest-assured:2.9.1-SNAPSHOT`) is just the bare-module.  Only the sizes and sha1sums give the game away, until you inspect the contents.

This might also suggest that the dist/full all-in-one jar might also be useful to push to Maven with some qualifier.

For now, I thought maybe add a qualifier/classifier to the dist uber-jar in the build, so it can no longer be confused with the module jar.

javabrett:
…dded -full qualifier to combined jar. Fixes #661.

Signed-off-by: Brett Randall javabrett@gmail.com

### DIFF
--- a/dist/dist-rest-assured/pom.xml
+++ b/dist/dist-rest-assured/pom.xml
@@ -42,7 +42,7 @@
                         </goals>
                         <configuration>
                             <appendAssemblyId>false</appendAssemblyId>
-                            <finalName>rest-assured-${project.version}</finalName>
+                            <finalName>rest-assured-full-${project.version}</finalName>
                             <descriptors>
                                 <descriptor>src/main/assembly/rest-assured-full-jar.xml</descriptor>
                             </descriptors>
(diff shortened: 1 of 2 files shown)


johanhaleby:
yeah this might be a good idea. Thanks for the pull request, I'll need to think about it a little bit (and my train has just arrived for work so need to go :))

PR #667, opened 2016-04-04
Improve Cookie Max-Age

TimonDL:
In [https://tools.ietf.org/html/rfc6265#page-20](specification) written:

> If delta-seconds is less than or equal to zero (0), let expiry-time
   be the earliest representable date and time.  Otherwise, let the
   expiry-time be the current date and time plus delta-seconds seconds.

This means that the value of Max-Age can be negative. In this case, browsers should set early represented date and time
I left an exception when the Max-Age < -1 because all the other cases do not make sense and is interpreted in the same way as -1. If you set Max-Age=-89219 it looks like a bug (I doubt at this point)

Correct me if I am wrong anywhere.

https://github.com/jayway/rest-assured/issues/580


johanhaleby:
Thanks for helping out and looking this up.

Issue #671, opened 2016-04-08
Suppport for setting session attributes in MockMvcRequestSpecification


sneyyar:
SpringMockMvc supports adding custom objects to session before sending the request.  This is critical to test Controllers implementing session management. It would be really helpful to add sessionAttribute(name, Object) to MockMvcRequestSpecification interface and link with corresponding feature supported by Spring mock mvc

sneyyar:
I already have this fixed to my local clone . Would be more than happy to check this in to master branch.

johanhaleby:
It would be great if you want to provide a pull request for this.

sneyyar:
I created a pull request with my fix. https://github.com/jayway/rest-assured/pull/673

johanhaleby:
I did some changes to make session attributes behave the same way as the other parameters/attributes. Thanks for your help-

sneyyar:
You are welcome. It would be nice if we could read objects (eg: userInfo)
from session after a get or post request. We need to inject userInfo into
session (session.sessionAttr(userInfo)  ) before invoking the next request.

Do we support this feature as of now ? If not, would you like to open
another issue ?

johanhaleby:
Perhaps you vould open another issue so that we can discuss it there?

sneyyar:
Sure. I can open another issue if springMvc supports ability to retrieve
objects from session.

PR #673, opened 2016-04-12
Suppport for setting session attributes in MockMvcRequestSpecification

sneyyar:
#671 
Enhanced RestAssured mockMvc to attach custom objects to session before sending the request


johanhaleby:
Thanks for helping out, really appreciated.

PR #675, opened 2016-04-15
CookieFilter

stravag:
The cookie filter can be used to keep track of all the cookies sent by the server and use them in subsequent requests. It might come in handy when more than just the SessionFilter is needed.


stravag:
Lol, I just realized that my test is wrong... will update PR

johanhaleby:
Nice initiative, I'm sure this could be handy as well. Thanks for the PR.

PR #695, opened 2016-05-18
Feature Request to add BigInteger NumberReturnType support for JsonPa…

JDKoder:
In order to validate any math integers in a json response, you currently need to know the
type ahead of time.  This is not possible if the range extends beyond java.lang.Integer or java.lang.Long.  If someone needs a consistent integer type across their assertions they may configure the JsonPathConfig with the NumberReturnType.BIG_INTEGER.  Since this is a catch all for integers of any size to prevent folding, it should be appropriate for asserting values on numbers that could have a range from very small to extremely large and are not known ahead of time.


johanhaleby:
Thanks for the pull request. I agree that something like this would be a good thing. But I guess that this change is not backward compatible with the old version?!

JDKoder:
How so?  It should still work for existing code.  Is there something that
iterates over that enum somewhere that I didn''t catch?

johanhaleby:
I couldn't see anything that would break backward compatibility but I just wanted to ask you as well. But now that I've looked more closely I can see that it shouldn't break compatibility. Thanks for the PR!

Issue #720, opened 2016-07-15
Cookie attributes are sent in request


dzieciou:
When trying to send a cookie with attributes, REST-assured includes its attributed in HTTP request:

``` java
List<Cookie> cookies = new ArrayList<>();
cookies.add(new Cookie.Builder("token", "tokenValue").setDomain("testing.com").setPath("/access").build());

given()
    .cookies(new Cookies(cookies))
    .get("/access")
```

will result in HTTP GET Request with

```
Cookie: token=tokenValue;Path=/access;Domain=testing.com
```

However, [RFC6265](https://tools.ietf.org/html/rfc6265#section-5.4) is explicitly saying to include cookie name and value in Cookie header. It does not mention attributes to be included. This makes sense, because Cookies attributes are there only for the client to decide whether a cookie should be included in the request or not.

`RequestSpecificationImpl.groovy` is responsible for that:

``` java
if (cookies.exist()) {
  http.getHeaders() << [Cookie: cookies.collect { it.toString() }.join("; ")]
}
```

where `Cookie#toString` is serializing not only cookie's name and value but also its attributed.

I will provide a pull request.

dzieciou:
@johanhaleby I can fix it but do you have some tests already that verify integration between REST-assured RequestSpecificationImpl and Apache HTTP Client? Basically, the tests that verify that REST-Assured requests are properly translated into HTTP Requests

dzieciou:
I guess we can close it now as you merged to master.

PR #722, opened 2016-07-15
#720 

dzieciou:
This is a fix for https://github.com/rest-assured/rest-assured/issues/720

Do you have some tests already that verify integration between REST-assured `RequestSpecificationImpl` and Apache HTTP Client? Basically, the tests that verify that REST-Assured requests are properly translated into HTTP Requests.


johanhaleby:
Thanks for spotting this and especially for the PR 👍

Issue #717, opened 2016-07-08
Mismatch description from TypeSafeDiagnosingMatcher is ignored


dzieciou:
Consider the following 2 tests
```java
RestAssured.given().get("http://www.bbc.co.uk").then().cookie("BBC-UID", IsNumberMatcher.isNumber());

assertThat("856767dfe9eb165074aa4af7317a8189babf2e0c6704c4a69a1097f21d5452340Apache-HttpClient/4.5.2%20(Java/1.8.0_45)", IsNumberMatcher.isNumber());
```

where the matcher is defined as follows (apologies for abstract example):
```java
    static class IsNumberMatcher extends TypeSafeDiagnosingMatcher<String> {

        @Override
        protected boolean matchesSafely(String actualNumber, Description mismatchDescription) {

            try {
                Integer.parseInt(actualNumber);
                return true;
            } catch (NumberFormatException e) {
                mismatchDescription.appendText(String.format("<%s> is some strange string", actualNumber));
                return false;
            }
        }

        public void describeTo(Description description) {
            description.appendText("small number");
        }

        public static IsNumberMatcher isNumber() {
            return new IsNumberMatcher();
        }
    }
```

The first test will print the following message when failing:
> Expected cookie "BBC-UID" was not small number, was "65e7d78f798c0312f48f6c43311aedb9e4252ee477246466fa40d742ed9482940Apache-HttpClient/4.5.2%20(Java/1.8.0_45)".

The second: 
> java.lang.AssertionError: 
Expected: small number
     but: <856767dfe9eb165074aa4af7317a8189babf2e0c6704c4a69a1097f21d5452340Apache-HttpClient/4.5.2%20(Java/1.8.0_45)> is some strange string

I would expect RestAssured not to ignore `mismatchDescription` of a custom matcher.

I tried with REST-Assured 3.0.0.

johanhaleby:
Sorry for the late response (I'm on vacation). REST Assured should take the `mismatchDescription` into account as well. Probably something I've missed. Feel free to send a pull request if you want to help out.

PR #723, opened 2016-07-15
#717

dzieciou:
Fix for issue #717: Mismatch description from TypeSafeDiagnosingMatcher is ignored.

The fix is in groovy and test the test is in Java, so it does not read nicely. Let me know if you want to get it improved.


johanhaleby:
Thanks a lot once again for PR (and for good test cases!) Great work!

Issue #726, opened 2016-07-22
JsonPath root path followed by array indexing


tobias-johansson:
I would like to do something like this (using `JSON` string from JsonPathTest.java):
```java
@Test
public void rootPathFollowedByArrayIndexing() throws Exception {
    final JsonPath jsonPath = new JsonPath(JSON).setRoot("store.book");
    assertThat(jsonPath.getString("[0].author"), equalTo("Nigel Rees"));
}
```
But this yields the following error:
```
java.lang.IllegalArgumentException: Invalid JSON expression:
Script1.groovy: 1: unexpected token: [ @ line 1, column 38.
                            store.book.[0].author
                                        ^
```
PR coming up :)

PR #727, opened 2016-07-22
Don't append '.' to root path if given path starts with array indexing

tobias-johansson:
Should solve issue #726
Can't think of any problems this might introduce...


tobias-johansson:
Hmm I don't think the test fails because of my changes?

johanhaleby:
Thanks for the PR. I don't think the failing test is due to this commit. Sorry for the confusion.

PR #746, opened 2016-09-22
ResponseSpecification: Don't join root and path with '.' if path starts with array indexing

feshbach:
This is the same bug that was fixed in #727 except for ResponseSpecificationImpl.


johanhaleby:
Thanks for the PR! Could you provide a test case as well though?

feshbach:
Sure. Where should it go? Do I need to create a new test class? Do I need to do anything to get the test to run?

johanhaleby:
There's class called `RootPathITest`, perhaps you could add it there?

johanhaleby:
Thanks a lot for your help!

PR #754, opened 2016-10-14
Allow keystore and truststore passwords to differ

mike42:
This pull request fixes a typo in `SSLConfig.java` which was included in commit a0adf58ba907ceaf5fe6076ea00589cc5c9c4b7e (splitting of keystore and trust store).

In the current `master` branch, when have an SSLConfig object and attempt to retrieve the trust store password, the key store password is returned intead.

This means it's not possible to use a keystore and truststore with different passwords. You will get a "Keystore was tampered with, or password incorrect" error if you attempt to do so, because the configured trust store password is never used.

This is not picked up in tests, as "test1234" is used as a password with some consistency.


johanhaleby:
Thanks for spotting this and for providing a PR!

PR #756, opened 2016-10-19
SSLConfig correction

mike42:
This pull request is another fix to the `SSLConfig` class. Currently, `getKeyStore()` is returning `trustStore`, rather than `keyStore`.

Related:
- #754


johanhaleby:
Thanks, and sorry for the typo.. That's what happens when you don't have enough tests.

Issue #758, opened 2016-10-24
Cookie domain not set for Set-Cookie without explicit domain attribute


dzieciou:
Hi,

When receiving HTTP Response with Set-Cookie header that does not have have explicit domain attribute, e.g.
```
    Set-Cookie: TOKEN=some value; Path=/; Secure; HttpOnly 
```
REST-assured 4.0.01 is NOT setting a default domain for cookie, i.e.,

```java
    given().
        baseUri("http://somedomain.com")
    when().
        ...
    then().extract().detailedCookies().get("TOKEN").getDomain()
```

will return null, while I would expect it to return a default domain, here ".somedomain.com", at least this is how for instance Apache HTTP Client and Web browsers behave.

The only reason REST-assured is not doing this is it was created to map HTTP responses as they are without any post-processing.

More details about domain calculation. A default domain is a domain guessed from HTTP request origin, precisely from Host header value. Apache HTTP Client calculates this default domain. If you look into org.apache.http.impl.cookie.BrowserCompatSpec (with my comments added to clarify)

```java
// Maciej: Setting default domain
cookie.setDomain(getDefaultDomain(origin));

// cycle through the parameters
final NameValuePair[] attribs = elem.getParameters();
for (int j = attribs.length - 1; j >= 0; j--) {
    final NameValuePair attrib = attribs[j];
    final String s = attrib.getName().toLowerCase(Locale.ROOT);
    cookie.setAttribute(s, attrib.getValue());
    final CookieAttributeHandler handler = findAttribHandler(s);
    // Maciej: For domain attribute finds RFC2965DomainAttributeHandler 
    if (handler != null) {
        // Maciej: RFC2965DomainAttributeHandler sets specific domain for a cookie if necessary
        handler.parse(cookie, attrib.getValue());
    }
}
```

johanhaleby:
Thanks for reporting

Lewka:
Hi there, is there any update about this issue? Cause I've faced the same situation, I get cookies without domain. To get cookies with domain I had to write code using HttpClient

johanhaleby:
@Lewka Please consider helping out with a PR

PR #759, opened 2016-10-25
Fix for #758: set default domainset for parsed cookies without specif…

dzieciou:
…ic domain

Fix for #758. It calculated default cookie domain based on origin host.

Please, review it carefully. I had a problem how to pass RequestSpecification instance so that implementations of ResponseSpecification are aware of it.

I am also not sure, if the feature I provide should be enabled by default. For instance, a user of REST-assured may want to check what actually is returned by the server, i.e., s/he may want to make sure a cookie is returned without specific domain. If this feature is enabled, calculated default domain will mask that.


dzieciou:
I can fix the failing part but let me  know first if you want this fix to be introduced at atll.

johanhaleby:
Thanks for the PR. I'll try to look into it on friday.

johanhaleby:
Again, thanks for supplying a PR. I agree that this should not be enabled by default since RA is a testing library and you should probably return what the server provides to you as default. 

But I wouldn't mind making this configurable as you say. One question though regarding the implementation. Wouldn't this be better implemented as a [filter](https://github.com/rest-assured/rest-assured/wiki/Usage#filters)? Or even better, wouldn't it be possible to just provide a configuration shortcut for configuring Apache HTTP Client to handle this for us?

PR #763, opened 2016-11-03
fix #762: merge all parameters correctly and set boundary if necessary

### LINKED ISSUE #762: unexpected behavious when using multipart/form-data
There are currently 2 problems when trying to use REST-Assured to produce multipart/form-data requests:
- form parameters (i.e. String-values) do not get merged correctly with the multipart parameters. so using reqSpec.formParam("key", "value") followed by reqSpec.multiPart(...) does not lead to the expected result.
- the boundary for the request has to be set explicitly in the "Content-Type" header of the request. otherwise it remains empty and will only be generated by the underlying MultipartEntityBuilder for the entity in the body.

I am currently preparing a PR which fixes these problems that i will post shortly

dornimaug:
this fixes the problems specified in https://github.com/rest-assured/rest-assured/issues/762

### DIFF
--- a/rest-assured/src/main/groovy/io/restassured/internal/RequestSpecificationImpl.groovy
+++ b/rest-assured/src/main/groovy/io/restassured/internal/RequestSpecificationImpl.groovy
@@ -1448,7 +1448,7 @@ class RequestSpecificationImpl implements FilterableRequestSpecification, Groovy
     }
     return mergeAndRemoveDoubleSlash(mergeAndRemoveDoubleSlash(baseUriPath, basePath), path)
   }
-
+  
   private def registerRestAssuredEncoders(HTTPBuilder http) {
     // Multipart form-data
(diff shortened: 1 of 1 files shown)


johanhaleby:
Thanks a lot for helping out. Before merging I just want to ask why the diff is so big that I can't even view it?

dornimaug:
I think i messed up the line endings. I'll look into it and fix it to the original so the diff will make sense. Sorry for the inconvenience.

johanhaleby:
No problem, would be great if you could address the formatting.

dornimaug:
I had to reset and reapply to get it to work but you can now see the actual changes here: https://github.com/rest-assured/rest-assured/pull/763/commits/f87bbbfbdf7eba258623c010c1ee93ebda33b4d0

johanhaleby:
Thanks!

PR #766, opened 2016-11-07
Fix 762 (PR#763) part 2

dornimaug:
I made a mistake with my last PR (https://github.com/rest-assured/rest-assured/pull/763) so the issue was not quite fixed by it. This is to correct it (and the line endings once again). The actual changes are in https://github.com/rest-assured/rest-assured/commit/774320565ceb7767a80c24ad3ef55c4a1e32f68e

Sorry once again for all the hassle


johanhaleby:
There seem to be a test failing. Is this due to your changes or could it be flaky!? (I've never seen it failing before though).

dornimaug:
Was a bit too eager there... I have travis working now on my fork repo so such mistakes won't happen in the future :) The last commit makes both scenarios (String and non-String values) working correctly now.

johanhaleby:
Great! Thanks!

PR #779, opened 2016-12-13
Fix wrong class name in the exception message

making:


johanhaleby:
Thanks for the PR! Really appreciated.

Issue #781, opened 2016-12-20
Mismatch description from Matcher is ignored for status code


jromero:
Trying to add additional information in regards to why the status code might have mismatched such as body using a custom Matcher but it's not possible due to RestAssured ignoring the mismatchDescription [here](https://github.com/rest-assured/rest-assured/blob/master/rest-assured/src/main/groovy/io/restassured/internal/ResponseSpecificationImpl.groovy#L528
). Similar to #717 

```
Expected status code <201> doesn't match actual status code <200>.
```

Matcher:
```java
final TypeSafeMatcher<Integer> matcher = new TypeSafeMatcher<Integer>() {
    @Override
    protected boolean matchesSafely(Integer integer) {
        return integer.equals(statusCode);
    }

    @Override
    public void describeTo(Description description) {
        description.appendValue(statusCode);
    }

    @Override
    protected void describeMismatchSafely(Integer item, Description mismatchDescription) {
        mismatchDescription.appendText("body: ").appendText(body);
    }
};
```

johanhaleby:
Thanks for the issue report. Perhaps you could help out and provide a pull request?

PR #782, opened 2016-12-21
#781: Fix: Mismatch description from Matcher is ignored for status code

jromero:
Created a reusable error message builder to enable consistent handling in the future and enable unit testing.

This PR only affects the error message produced when status code is mismatching to maintain scope.


johanhaleby:
Thanks a lot, really appreciated.

PR #784, opened 2017-01-07
Added ability to order filters

jcravi:
Please let me know if I need more tests (and where?)

### DIFF
--- a/rest-assured/src/main/java/io/restassured/RestAssured.java
+++ b/rest-assured/src/main/java/io/restassured/RestAssured.java
@@ -21,6 +21,7 @@ import io.restassured.builder.RequestSpecBuilder;
 import io.restassured.builder.ResponseSpecBuilder;
 import io.restassured.config.*;
 import io.restassured.filter.Filter;
+import io.restassured.filter.OrderedFilter;
 import io.restassured.filter.log.LogDetail;
 import io.restassured.http.Method;
 import io.restassured.internal.*;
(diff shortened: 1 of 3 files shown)


johanhaleby:
Thanks a lot for the PR! I'll try to review it soon.

johanhaleby:
Looks good to me! I'll probably add a bit more javadoc. Thanks a lot, good feature.

johanhaleby:
I've refactored this somewhat. Sorting is lazy for example, will that have any negative impact that you're aware of?

jcravi:
That's great! Thanks. The only negative I can think of right now is the performance, but I doubt that will be noticeable unless someone has upwards of a 100 filters. I would have preferred to create a sorted list, but since that doesn't exist, a treeset with a comparator. Also, if we could've have used java 8, default methods in interfaces.
I wanted to minimize what I was changing, since I don't fully understand the code yet.

johanhaleby:
Yeah I doubt that it'll have any noticeable performance impacts. It would be great if you could try out the latest snapshot (version `3.0.2-SNAPSHOT`) to verify that it works for you after having added to following maven repo:

```xml
<repositories>
        <repository>
            <id>sonatype</id>
            <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
            <snapshots />
        </repository>
</repositories>
```

jcravi:
I will check it out today.

jcravi:
Apologies for taking so long. I will try to get this done sooner rather than later.

jcravi:
Alright, I was able to test it out with our scenario and it worked perfectly. Thanks again for the quick turnaround @johanhaleby!

PR #801, opened 2017-02-03
fix #791: check cookies of HttpServletResponse (additionally to the Set-Cookie header)

### LINKED ISSUE #791: response cookies are not copied
I cannot access the cookies from the response:
```
Map<String,String> cookies = given()
                .when()
                .get("/?lang=de")
                .then()
                .statusCode(HttpStatus.OK.value())
                .extract().cookies();
```

The `cookies` map is always empty.

I traced the bug to the file https://github.com/rest-assured/rest-assured/blob/6b5fea6a281fd45e2bf19121a726c8ea1ec7b7a0/modules/spring-mock-mvc/src/main/java/io/restassured/module/mockmvc/internal/MockMvcRequestSenderImpl.java#L182 - it seems that the cookies from the `MockHttpServletResponse` object aren't copied into the `MockMvcRestAssuredResponseImpl` object.

andihit:
at the moment only the Set-Cookies header is examined for cookie validations. this PR fills the restAssuredResponse.cookies field and merges the cookies from the Set-Cookies header with the cookies from the httpServletResponse.getCookies() for validations (`CookieMatcher`).

### DIFF
--- a/modules/spring-mock-mvc/src/main/java/io/restassured/module/mockmvc/internal/MockMvcRequestSenderImpl.java
+++ b/modules/spring-mock-mvc/src/main/java/io/restassured/module/mockmvc/internal/MockMvcRequestSenderImpl.java
@@ -149,6 +149,27 @@ class MockMvcRequestSenderImpl implements MockMvcRequestSender, MockMvcRequestAs
         return new Headers(headers);
     }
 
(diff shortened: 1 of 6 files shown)


johanhaleby:
That's great!! Thanks a lot for helping out, really appreciated!

PR #807, opened 2017-02-10
Fix double negative in Exception message

MarkNBroadhead:
Remove second negative from exception message "... no supported Content-Type was not specified... ".


johanhaleby:
Thanks!

Issue #824, opened 2017-03-01
Add RequestSpecification get config


jromero:
The current implementation of RequestSpecification have a method `getConfig` which allows you to further manipulate the configuration later in the process. 

This method should be added to the interface but probably renamed to `.config()` for DSL purposes since it would be the preferred way to implement further implementations while still allowing the configuration to be modified at later stages.

Current use case:

BEFORE
```
@Override
public void beforeRequest(RequestSpecification requestSpecification) {
    RestAssuredConfig config;
    if (requestSpecification instanceof FilterableRequestSpecification) {
        config = ((FilterableRequestSpecification) requestSpecification).getConfig();
    } else {
        throw new IllegalArgumentException("Don't know how to retrieve current configuration from " + requestSpecification);
    }

    requestSpecification
        .config(config.logConfig(logConfig().defaultStream(LOG_STREAM)))
        .log().all();
}
```

AFTER (PROPOSAL)
```
@Override
public void beforeRequest(RequestSpecification requestSpecification) {
    requestSpecification
        .config(requestSpecification.config().logConfig(logConfig().defaultStream(LOG_STREAM)))
        .log().all();
}
```

PR #834, opened 2017-03-07
#824: Add RequestSpecification config getter

jromero:


johanhaleby:
Before accepting this I'd like to know more about your use case. Why is that you can't use a filter to get a hold off and alter the configuration?

jromero:
@johanhaleby I just tried creating a filter but it doesn't work in my use case because altering RequestSpecification log configuration within a filter causes a `ConcurrentModificationException`.

```
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at java_util_Iterator$next$0.call(Unknown Source)
	at io.restassured.internal.filter.FilterContextImpl.next(FilterContextImpl.groovy:70)
	at lv.ctco.cukesrest.internal.logging.HttpLoggingPlugin$LoggingConfigFilter.filter(HttpLoggingPlugin.java:166)
```

It's due to the fact that modifying the log config adds a filter (while looping filters):
```groovy
  private def logWith(LogDetail logDetail, boolean prettyPrintingEnabled, PrintStream printStream) {
    requestSpecification.filter(new RequestLoggingFilter(logDetail, prettyPrintingEnabled, printStream, shouldUrlEncodeRequestUri(requestSpecification)))
    requestSpecification
  }
```

So essentially it's too late in the process to alter the log configuration in a filter.

jromero:
To be as concise as possible, it's not really the log config that's causing the issue it's the log spec `RequestLogSpecification`.

```java
class LoggingConfigFilter implements OrderedFilter {

    @Override
    public Response filter(FilterableRequestSpecification filterableRequestSpecification, FilterableResponseSpecification filterableResponseSpecification, FilterContext filterContext) {
        final RestAssuredConfig config = filterableRequestSpecification.getConfig();
        config.logConfig(config.getLogConfig().defaultStream(logStream));

        final RequestLogSpecification logSpec = filterableRequestSpecification.log();
        final List<LogDetail> logDetails = Lists.newArrayList(LogDetail.ALL); // provided externally

        details:
        for (LogDetail detail : logDetails) {
            switch (detail) {
                case ALL:
                    logSpec.all();
                    break details;
                case BODY:
                    logSpec.body();
                    break;
                case COOKIES:
                    logSpec.cookies();
                    break;
                case HEADERS:
                    logSpec.headers();
                    break;
                case METHOD:
                    logSpec.method();
                    break;
                case PARAMS:
                    logSpec.parameters();
                    break;
                case URI:
                    logSpec.uri();
                    break;
            }
        }

        return filterContext.next(filterableRequestSpecification, filterableResponseSpecification);
    }

    @Override
    public int getOrder() {
        return HIGHEST_PRECEDENCE;
    }
}
```

johanhaleby:
Oh you're absolutely right. It looks like a bug to me that one should get a `ConcurrentModificationException`. What if this would be fixed? Would the filter approach work then?

jromero:
@johanhaleby If it gets fixed by means that I can then alter the `RequestLogSpecification` in a filter and it take immediate effect then yes I suppose that would provide a solution and this PR can be omitted.

johanhaleby:
But would you still say that it would be more convenient to have  the `config()` method regardless of a potential fix in the filter?

jromero:
From an ease of use perspective it would be extremely more convenient.
Specially since I didn't even know about filters. That said I do understand
if it's preferable to use filters if this PR conflicts with the design of
the library.

Issue #858, opened 2017-05-05
Feature request: take CharSequence rather than String in get(..), post(...) etc


weaselmetal:
We are testing our API that uses Katharsis framework with REST Assured.
The query parameter naming is a bit verbose, like ?filter[person][name]=tom.
So I created a little helper class, that allows me to construct a String representing the API-URL to test like this:

    String url = getApiResourceRoute()
                .withFilterParam("name", "Tom")
                .withFilterParam("age", 30)
                .toString();

producing:  `http://localhost:8080/api/persons?filter[persons][name]=Tom&filter[persons][age]=30`

The information that this request is about `persons` is stored in a field of the test class, in case you were wondering.

Since the `get(...)` methods and it's siblings only take types that are final, I'll always need to do something like `toString()` as my last method call in the chain. Now, if your `get(...)` methods would take a `CharSequence` instead of a `String`, I could also pass my fancy helper class directly, because it implements `CharSequence`. I would love to no not have the `toString()` there.

Would make my tests with Rest Assured even more beautiful to behold ;).

johanhaleby:
Interesting idea. I don't think it will hurt backward compatibility to change everything to CharSequence?

weaselmetal:
I guess it shouldn't. A `String` is a `CharSequence`. Most of the Apache `StringUtils` methods take a `CharSequence`.

johanhaleby:
Ok I'll give it a try

johanhaleby:
Ok that was quick try :) There are many places in Apache HTTP Client that only accept strings and not CharSequence. I think this will be too much work for what it's worth and it'll lead to casting in a lot of places. I'm sorry to say but think it's better for you to stick with `toString`.

weaselmetal:
Thanks for trying, though I am of course very very sad now.

For what it's worth, I hoped it could be as easy as (speaking Java)

before:

    get(String path)
    {
        doSomethingWithPath(path);
    }

after:

    get(CharSequence path)
    {
        doSomethingWithPath(path.toString());
    }

well, sort-of Java, minus bloat.

johanhaleby:
yeah, it's just that this would require a lot of changes. But if you provide a pull request that works I would probably reconsider ;)

weaselmetal:
Gonna check. Never put my hands on groovy before :)

weaselmetal:
[...]

PR #860, opened 2017-05-08
 #858: CharSequence rather than String

weaselmetal:
Here's my PR for #858


johanhaleby:
Oh it wasn't as hard as I expected. Thanks for doing this!

johanhaleby:
@weaselmetal It turns out that this commit is not binary compatible (only source compatible) so I was forced to revert it.

Since this change is (unfortunately) backward incompatible it'll probably have to wait until the next major release before we can merge it again.

weaselmetal:
SAD!On Oct 5, 2017 20:12, Johan Haleby <notifications@github.com> wrote:@weaselmetal It turns out that this commit is not binary compatible (only source compatible) so I was forced to revert it.
Since this change is (unfortunately) backward incompatible it'll probably have to wait until the next major release before we can merge it again.

johanhaleby:
Sorry :( Thanks for the effort anyways

weaselmetal:
No worries :). Keeping my `toString()` around ain't gonna kill me.

Issue #861, opened 2017-05-08
Assertion contentType(BINARY) always fail


gfouquet:
Assertions on content type with a ContentType.BINARY argument always fail with the message below.

```
    @Test 
    public void should_have_a_binary_content_type() {
        when().get("/my-binary-resource")
        .then().contentType(ContentType.BINARY);
    }

```

```
java.lang.AssertionError: 1 expectation failed.
Expected content-type "BINARY" doesn't match actual content-type "application/octet-stream; charset=ISO-8859-1".
```

I found the cause, I'll submit a PR shortly

johanhaleby:
Oh thanks for spotting this and double thanks for the PR.

gfouquet:
You're welcome, Rest assured is a helpful tool.

PR #862, opened 2017-05-08
Proposed fix for issue #861

gfouquet:
ContentType.fromContentType() did not correctly coerce some content types.

I added an integration test on `contentType(BINARY)` assertion and a unit test on `fromContentType` which should detect bugs when an enum is added to `ContentType`


johanhaleby:
Thanks a lot, looks great!

PR #864, opened 2017-05-12
Upgrading Apache HTTP Client to v4.5.3

rytighe:
In using RA on my current project, I've noticed an issue with two-way SSL and a seeming root cause with Apache HTTP Client v4.5.2.  Upgrade to 4.5.3 works perfectly.  

There do seem to be some tickets in the HTTP Client changelog of note - a bit over my head, though: https://archive.apache.org/dist/httpcomponents/httpclient/RELEASE_NOTES-4.5.x.txt


johanhaleby:
Thanks!

PR #869, opened 2017-06-01
ntlm authentication

paweld2:
DSL changes to allow the configuration of NTLM authentication on the httpclient.


johanhaleby:
Thanks a lot!! I'll look into this soon!

Issue #814, opened 2017-02-20
EOFException occurs when POST response have an empty body with gzip content type


anatoly-antonov:
When a POST request returns with empty body and Content-Encoding is equal to gzip, an EOFException exception is thrown. See exception stack trace below:

```java
Caused by: java.io.EOFException
  at java.util.zip.GZIPInputStream.readUByte(GZIPInputStream.java:268)
  at java.util.zip.GZIPInputStream.readUShort(GZIPInputStream.java:258)
  at java.util.zip.GZIPInputStream.readHeader(GZIPInputStream.java:164)
  at java.util.zip.GZIPInputStream.<init>(GZIPInputStream.java:79)
  at java.util.zip.GZIPInputStream.<init>(GZIPInputStream.java:91)
  at com.jayway.restassured.internal.http.GZIPEncoding$GZIPDecompressingEntity.getContent(GZIPEncoding.java:66)
  at org.apache.http.conn.BasicManagedEntity.getContent(BasicManagedEntity.java:87)
  at com.jayway.restassured.internal.http.HTTPBuilder.parseResponse(HTTPBuilder.java:544)
  at com.jayway.restassured.internal.RequestSpecificationImpl$RestAssuredHttpBuilder.super$2$parseResponse(RequestSpecificationImpl.groovy)
  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:497)
  at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:93)
  at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
  at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1210)
  at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.invokeMethodOnSuperN(ScriptBytecodeAdapter.java:132)
  at com.jayway.restassured.internal.RequestSpecificationImpl$RestAssuredHttpBuilder.parseResponse(RequestSpecificationImpl.groovy:2056)
  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #874, opened 2017-06-19
Adjustments for #814

sergeytrasko:
Handle situation when response is GZIP-ed, but it's empty


johanhaleby:
Thanks for the PR. However I'm hesitant to bring in an additional library (commons-io) just for reading an inputstream inte a byte-array. I'd rather just have the method copied or implemented manually.

johanhaleby:
Excellent work! Thanks a lot for helping out, really appreciated. And extra points for writing tests!

PR #890, opened 2017-07-24
treat text/json as JSON content type

archeus:
We have a legacy application that is returning JSON but declares the content type as `text/json`. 
Although `text/json` is not a standard MIME type, adding support `ContentType.JSON` would make the code consistent - the [Parser](https://github.com/rest-assured/rest-assured/blob/7cc7503b66d17fffa7d6f8442e3838a9b0d5f7b8/rest-assured/src/main/java/io/restassured/parsing/Parser.java#L29) is already taking it into account, while [ContentType](https://github.com/rest-assured/rest-assured/blob/13b5af0822a3366dcc67f5beeb22e8fdb3eafc1a/rest-assured/src/main/java/io/restassured/http/ContentType.java#L69) does not.

See also:
* https://stackoverflow.com/questions/477816/what-is-the-correct-json-content-type#answer-2590013


johanhaleby:
Thanks!

PR #894, opened 2017-08-04
practice for git tfs

avinashkandukuri:


johanhaleby:
Thanks for the effort but I'm going to reject this PR since REST Assured is using Git and not TFS and I don't like to "pollute" the project with tfs specific files.

Issue #912, opened 2017-09-18
Set authentication once to set in SecurityContextHolder and Principal in method parameter


lsiu:
Once we set the authentication, for example: 

https://github.com/rest-assured/rest-assured/blob/d3a34e9a6bb72cd37f509e1dc12eafbd66642b71/modules/spring-mock-mvc/src/test/java/io/restassured/module/mockmvc/SecuredControllerTest.java#L72-L75

We should be able to retrieve it through:

1. SecurityContextHolder.getContext().getAuthentication()
2. @GetMapping public void someMethod(Principal principal) 

no. 2 should automatically apply since Authentication extends Principal

Looking at the code, it appears we just need to change the "else if" to a separate "if" statement.

https://github.com/rest-assured/rest-assured/blob/d3a34e9a6bb72cd37f509e1dc12eafbd66642b71/modules/spring-mock-mvc/src/main/java/io/restassured/module/mockmvc/internal/MockMvcRequestSenderImpl.java#L181-L185

I can create a pull request and update some test, but the [SecuredControllerTest](https://github.com/rest-assured/rest-assured/blob/master/modules/spring-mock-mvc/src/test/java/io/restassured/module/mockmvc/SecuredControllerTest.java) seems to separate this out deliberately?

Perhaps @johanhaleby can comment?

lsiu:
Better illustrate with a PR. Closing this issue in favor of PR #913.

PR #913, opened 2017-09-18
Fix for #912, Set authentication once to set in SecurityContextHolder and Principal in method parameter

lsiu:
Once we set the authentication, for example: 

https://github.com/rest-assured/rest-assured/blob/d3a34e9a6bb72cd37f509e1dc12eafbd66642b71/modules/spring-mock-mvc/src/test/java/io/restassured/module/mockmvc/SecuredControllerTest.java#L72-L75

We should be able to retrieve it through:

1. SecurityContextHolder.getContext().getAuthentication()
2. @GetMapping public void someMethod(Principal principal) 

no. 2 should automatically apply since Authentication extends Principal

Looking at the code, it appears we just need to change the "else if" to a separate "if" statement.

https://github.com/rest-assured/rest-assured/blob/d3a34e9a6bb72cd37f509e1dc12eafbd66642b71/modules/spring-mock-mvc/src/main/java/io/restassured/module/mockmvc/internal/MockMvcRequestSenderImpl.java#L181-L185

I can create a pull request and update some test, but the [SecuredControllerTest](https://github.com/rest-assured/rest-assured/blob/master/modules/spring-mock-mvc/src/test/java/io/restassured/module/mockmvc/SecuredControllerTest.java) seems to separate this out deliberately?

Perhaps @johanhaleby can comment?


johanhaleby:
Thanks for your help!

Issue #956, opened 2017-12-21
Add cookie filter with cookie attributes matching.


MaciejCiszewski:
Currently we do have CookieFilter that parses the cookies from the response and adds them to the request unconditionally in terms of the domain or other cookie attributes like path etc. 
Wouldn't it be useful to create more advanced version of this filter that does proper cookie matching against the request URL so that it works more like a real browser?

johanhaleby:
Yeah that sounds like a good idea. Maybe you could help out?

MaciejCiszewski:
@johanhaleby Sure, PR added: [https://github.com/rest-assured/rest-assured/pull/957](url)

PR #957, opened 2017-12-21
#956 - Cookie matching filter

MaciejCiszewski:


johanhaleby:
Big thanks for the PR. Just one thought, what do you think about merging this functionality into the existing CookieFilter instead of introducing a new filter? Maybe it's more confusing to have 2 filters than to have one to which you can provide some settings or configuration?

MaciejCiszewski:
I agree, It was hard for me to decide which way to go but your suggestion is fine to me. I will update PR.

johanhaleby:
Thanks for the update. From what I see you've pretty much replaced the original CookieFilter with your updated version. Is this always what you want? Is there no use case for copying all cookies that the current filter is doing?

MaciejCiszewski:
Copying all cookies regardless of the domain is not what RFC6265  - (HTTP State Management Mechanism) specifies and not the way browsers work. Won't the testing with use of this filter be more accurate from now? Copying all the cookies would simulate the browser or client that in general misbehaves according to the standards.

johanhaleby:
Ok that sounds like a valid reason to me. If people complain then maybe we have to add some additional configuration options but for now I think this is better. It'll be backward incompatible but we could probably argue that the previous implementation was buggy. 

So thanks for you help, really appreciated.

PR #966, opened 2018-01-19
Test cleanup: WithJetty should be an abstract class

alb-i986:
Then, no need to @Ignore it.


johanhaleby:
True, I didn't know that at the time! Thanks for the fix.

PR #971, opened 2018-01-31
Fix os dependant code

mmitteregger:
This should fix "mvn clean install" on Windows.
I have separated the changes into 4 commits of which one changes the produced log output (request/response printing) and the other 3 only touch test code.

1.  Consistently use the system line separator for request/response printing
Previously the printed request and response output was a mix between new lines and the system line separator. Without this change it was impractical to fix the tests and I believe it should be fixed anyway.

2. Specify file encoding for maven-surefire-plugin
This is required for test files that contain non ASCII characters.
More details about this can be found in a [StackOverflow post](https://stackoverflow.com/questions/17656475/maven-source-encoding-in-utf-8-not-working).

3. Fix os dependant tests because of different line separators
Should be pretty self-explanatory.
I have opted to use String.format() with the %n format specifier to be os independant instead of using String concatenation with the system line separator.

4. Disable ResponseTimeITest on Windows
Some of these tests fail when doing a complete "mvn clean install" run, because the response time was actually 0. They always passed when the tests were run standalone.
I think the reason for this is because the precision of System.currentTimeMillis() is lower on Windows and the tests are actually too fast with a warm JVM. I'd like to conditionally disable them on Windows for now.


johanhaleby:
Thanks a lot!! I'll try to look into the PR tomorrow.

johanhaleby:
@mmitteregger I'm terribly sorry but I have completely forgotten about this PR! I know this is probably too much to ask but could you try to resolve the conflicts and I'll accept it asap.

mmitteregger:
@johanhaleby thanks for still looking to get this merged.
I have rebased the commits on top of the current master branch.
Apart from resolving the merge conflicts and modifying additional tests there is now an additional 5th commit.

**Clean up test log file and directory with File#deleteOnExit()**
The LoggingToDiskTest created a temporary directory and file. One test method then tried to recursively delete the directory, but the WriteLogsToDisk rule still had on open OutputStream.
This always failed on my Windows machine, because a file cannot be deleted if there still is an open file handle.
I have opted to delete the files with [File#deleteOnExit](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--) which basically defers the deletion to JVM exit with a shutdown hook.

Please let me know if you are ok with the implementation details, especially:
1. Using `SystemUtils.LINE_SEPARATOR`
2. Using `String.format(...)` in the test code
3. Using `File.deleteOnExit` for the appropriate test file cleanup

PR #972, opened 2018-02-05
Detailed cookie matcher introduced.

rafalsiwiec:
This change introduces ability for detailed cookies assertions.

### DIFF
--- a/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/CookieITest.java
+++ b/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/CookieITest.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 
 import static io.restassured.RestAssured.*;
+import static io.restassured.matcher.RestAssuredMatchers.detailedCookie;
 import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.assertThat;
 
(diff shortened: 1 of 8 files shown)


johanhaleby:
Nice! Thanks a lot for the contribution. But would it be possible to add an overloaded method instead of introducing a new one with different name? I.e. instead of:

```java
given().
       get("/multiCookie").
then().
      detailedCookie("cookie1", detailedCookie().maxAge(1234567));
```

it might be possible to do:

```java
given().
       get("/multiCookie").
then().
      cookie("cookie1", detailedCookie().maxAge(1234567));
```

Maybe it'll be easier to find and more intuitive for people than `detailedCookie`?

WDYT?

rafalsiwiec:
@johanhaleby since:

```public T cookie(String cookieName, Matcher<?> expectedValueMatcher)```

method exists, we would need to be more specific and use exact matcher type - `DetailedCookieMatcher` - in the method signature.

So, instead of:

```T cookie(String cookieName, Matcher<? super Cookie> detailedCookieMatcher)```

it would be:

```T cookie(String cookieName, DetailedCookieMatcher detailedCookieMatcher)```

Initially, I thought that flexibility to use any matcher you want, for detailed cookie assertions would be useful. But since the case is rather narrow, and `DetailedCookieMatcher` provides fluent api for all properties I agree with your point.

johanhaleby:
Thanks for the contribution, highly appreciated!

rafalsiwiec:
Thanks for accepting this @johanhaleby :)
When do you plan to release next version?

johanhaleby:
Dunno. There's so much stuff to do and fix but I'm short on time. Maybe I should just release a new point release soon.

PR #976, opened 2018-02-09
DetailedCookieMatcher generics fixed

rafalsiwiec:
@johanhaleby it's shameful, but it looks like I messed up generics and `DetailedCookieMatcher` is not really usable with matchers right now.
This change should make it better.

BTW, are there any particular issues I could help you with? I'm keen to contribute but I don't know which things are most important.


johanhaleby:
No problem, it happens :)

> BTW, are there any particular issues I could help you with? I'm keen to contribute but I don't know which things are most important.

I would be extremely happy to  receive help implementing or fixing some issues. I would suggest that you just pick something that sounds interesting to you or something that you might fancy using yourself. Just ping me if you need any guidance or help.

Issue #974, opened 2018-02-08
Content-Length header should not be present in GET request but it is set to 0


Bisk1:
According to RFC 7230, chapter 3.3.2:

> A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body.

but when sending GET, this header is set to 0, which causes some servers interpret it incorrectly.

After some debugging, I found out that this is because HttpRequestFactory class does not have GET method in internal map so GET is represented as CustomHttpMethod which is subclass HttpEntityEnclosingRequestBase (and GET does not enclose entity). Apache HTTP library adds the Content-Length header for all instances of this class.

The map only has PUT, POST and PATCH. I guess that it should be populated with all other standard HTTP methods, for which implementations exist anyway in org.apache.http.client.methods package along those already added.

Bisk1:
Fixed in PR #977

PR #977, opened 2018-02-10
Add proper support for standard HTTP methods (#974)

Bisk1:


johanhaleby:
Thanks! There are some tests that are failing though. Is this expected with your change?

Bisk1:
@johanhaleby No, I will check it

Bisk1:
@johanhaleby  This is my second attempt -  I changed 3 public methods of HTTPBuilder - is it part of public API?

johanhaleby:
> @johanhaleby This is my second attempt - I changed signature of 3 public methods of HTTPBuilder - is it part of public API?

No it shouldn't be (since it's located in the "internal" package).

Thanks again for the update. I'll try to look into it when I have to opportunity.

johanhaleby:
I've look into the PR now and it looks good as far as I can tell but I still need to understand what problems we're solving here. Is it https://github.com/rest-assured/rest-assured/issues/974 specifically or does this solves even more problems? 

And you're right that it's by design that REST Assured supports sending body's in for all HTTP methods (including GET) since many people have asked for it for testing purposes.

johanhaleby:
That's great! Thanks for the clarification and big thanks for the pull request and for helping out!

PR #980, opened 2018-02-12
[Feature] Type instead of Class has been added for mapping

viclovsky:
Added `.as()` method to map any `Type`. It is more general solution than `Class<T>` so now we are able to use Gson, Jackson and Jackson2 by default which was not supported by Rest-assured. 
Added necessary tests.

### DIFF
--- a/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/CustomObjectMappingITest.java
+++ b/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/CustomObjectMappingITest.java
@@ -31,6 +31,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.lang.reflect.Type;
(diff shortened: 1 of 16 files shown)


johanhaleby:
This is great stuff, thanks a lot! And it should be backward compatible as well right?

viclovsky:
Yep, it backward compatible should be as well.
I tested this on my projects.

viclovsky:
Without the PR I can't customize mapping of my own http client based on Rest-assured https://github.com/swagger-api/swagger-codegen/pull/7492/files#diff-4ed6d557101bb6b37c06fbc3f44572b1R117 . Rest-assured client in swagger:  https://github.com/swagger-api/swagger-codegen/pull/7492 
Mapping in swagger is complex and it operates with Type a lot — it based on it.
Now I don't have any idea how to keep backward compatibility.

johanhaleby:
Thanks for your comment, I think it's an ok trade-off since it shouldn't be something that most people are using and it should be easy to update. Really cool that you're working on  swagger integration.

And thanks for the PR!

Issue #979, opened 2018-02-12
Resetting preemptive auth doesn't work


niriak:
Trying to reset a preemptive authentication entry doesn't work.

Example:
RestAssured.given().auth().preemptive().basic("username","password").auth().none()

Expected: No authentication information is send to the server
Actual: The basic authentication information is send to the server

Found on version 3.0.6

Looks like the authentication information is written into the header field "Authorization" and "auth().none()" doesn't remove this header entry.

johanhaleby:
oh that looks like a bug indeed! Thanks for reporting. It would be even nicer if you could provide a pull request :)

PR #994, opened 2018-04-01
Fix for #979 Removing Authorization header when setting auth().none()

jovanovicivan:
I don't know if this is the most generic way to do it but seams to work. This is the only auth header that I noticed so I guess it's safe to do it like this. If you have any other suggestions I'm open for discussion.


johanhaleby:
Thanks! A test would be nice, but otherwise I think it's ok

jovanovicivan:
Great, I would be very open to writing some tests as well, but I couldn't figure out how to run those in rest-assured-itest-java module. They all failed for me, so I tested manually with 4-5 test cases. If you have some pointers on how to run those tests that would be great, if not I'll take a look at them but probably in two weeks.

PR #996, opened 2018-04-04
Change firstName to lastName in Javadoc

rohitkadam19:
Changed firstName to lastName in java doc.


johanhaleby:
Thanks for noticing and thanks for PR!

PR #1013, opened 2018-05-11
Ensure RequestSpecBuilder picks up filters from static config (#1012)

### LINKED ISSUE #1012: Global filters not applied when using given(request, response)
Given the following test:

```java
    @BeforeClass
    public void setup() {
        RestAssured.filters(new RequestLoggingFilter(System.out));
    }

    @Test
    public void hello() {
        RequestSpecification requestSpec = new RequestSpecBuilder()
                .addHeader("hello", "world")
                .build();
        ResponseSpecification responseSpec = new ResponseSpecBuilder().expectStatusCode(401).build();
        given(requestSpec, responseSpec).get("https://example.com");
    }
```

logging will not occur, because the requestSpecification will never pick up the global filter value. Using given().spec(request) on the other hand works.

aldaris:
This PR should resolve #1012 by making sure that filters are passed to RequestSpecificationImpl upon creation.

### DIFF
--- a/rest-assured/src/main/java/io/restassured/builder/RequestSpecBuilder.java
+++ b/rest-assured/src/main/java/io/restassured/builder/RequestSpecBuilder.java
@@ -76,8 +76,8 @@ public class RequestSpecBuilder {
     private RequestSpecificationImpl spec;
 
     public RequestSpecBuilder() {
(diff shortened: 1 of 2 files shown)


aldaris:
Looks like static headers aren't picked up either, I'll have a look at that too, as it seems to originate from the same place.

johanhaleby:
Thanks a lot for the PR! Very appreciated.

PR #1017, opened 2018-05-17
Support multiple 'Set-Cookie' headers in a response using CookieFilter

runtarinn:


johanhaleby:
Thanks a lot for the PR, really appreciated!

Issue #1014, opened 2018-05-14
ResponseSpecBuilder can't be used with DetailedCookieMatcher


aldaris:
The expectCookie methods on ResponseSpecBuilder only allow Matcher<String> arguments, whilst ResponseSpecificationImpl exposes DetailedCookieMatcher as well.
The not so pretty workaround is to do this:

```java
        return new ResponseSpecBuilder()
            .expectStatusCode(200)
            .build().cookie("iPlanetDirectoryPro", detailedCookie().maxAge(-1));
    }
```

johanhaleby:
Oh that's true. Good find and thanks for pointing out. Care to provide a pull request?

aldaris:
Yepp, should be easy to do :)

PR #1018, opened 2018-05-18
Add detailed cookie matching to RequestSpecBuilder (#1014)

aldaris:
I've added the missing cookie method to ResponseSpecBuilder and updated the JavaDoc on ResponseSpecification to match reality.


johanhaleby:
Thanks a lot for all your help @aldaris. Really appreciated!

PR #1020, opened 2018-05-19
Fixed OAuth 1 for some methods

dlenroc:
When GET request with content-type **application/x-www-form-urlencoded** is made I get
```ava
java.lang.ClassCastException: org.apache.http.impl.client.RequestWrapper cannot be cast to org.apache.http.impl.client.EntityEnclosingRequestWrapper
```
HTTPRequest is an instance of EntityEnclosingRequestWrapper only when body or form-param is sending

also, `URLEncodedUtils.parse(entity);` can return NameValuePair with null value (instead of empty string)


johanhaleby:
Thanks a lot!

PR #1021, opened 2018-05-20
Add unit tests for ResponseSpecBuilder

Bisk1:
I thought the project would benefit from better coverage, so these are just tests to some previously untested code.


johanhaleby:
Awesome @Bisk1! Thanks a lot!

Issue #1040, opened 2018-07-17
Supplying InputStream as body() with content type of JSON tries to serialize InputStream object


m4ksio:
Trying to send a JSON read from InputStream, like
    
    given().
            contentType(ContentType.JSON).
            body(getClass().getResourceAsStream("/payment-1.json")).
    
results in

    Internal error: Can't encode java.io.BufferedInputStream@7cfdc0c2 to JSON.
        java.lang.UnsupportedOperationException: Internal error: Can't encode java.io.BufferedInputStream@7cfdc0c2 to JSON.

Looks like it tries to serialize the InputStream object itself instead of reading it directly.

johanhaleby:
What HTTP verb are you using?

johanhaleby:
Fixed in PR!

honza-zidek:
> Fixed in PR!

I am using rest-assured 4.3.3 and still the same error.

My code:

```
class MyTest  {
    @Value("classpath:/valid-request.xml")
    private Resource validRequest;
    ....
    @Test
    void myTest() {
        given()
                .header("Content-Type", "text/xml; charset=UTF-8")
                .header("SOAPAction", "myService")
                .body(validRequest.getInputStream())  // this does not work
//              .body(validRequest.getFile())  // this works
        .when()
                .post(PATH)
        .then()
                .log().ifValidationFails()
                .statusCode(200)
                .contentType("text/xml")
        ;
        
    }
}
```

However, rest-assured interprets the `InputStream` as an Object and cannot handle it.
When I replace `getInputStream()` with `getFile()`, it works.

PR #1042, opened 2018-07-18
Properly encodes JSON from InputStream. Fixed [#1040]

m4ksio:


johanhaleby:
Looks like there's a problem with some tests but they don't seem to be related to your PR!

So thanks a lot for providing a pull request! Really appreciated.

PR #1061, opened 2018-09-07
Make it clear that junit is only a test scope dependency

epabst:
Only use maven dependencyManagement to control the version
to avoid unintended or implicit effects.
The <optional>true</optional>'s are duplicated.

### DIFF
--- a/examples/rest-assured-itest-java/pom.xml
+++ b/examples/rest-assured-itest-java/pom.xml
@@ -72,6 +72,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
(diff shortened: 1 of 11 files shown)


johanhaleby:
Thanks but I've quite concerned about the removal of the optional dependencies. They are indeed optional by design, it shouldn't be required by every user of REST Assured to have three different JSON parsing libraries in the classpath for example.

epabst:
I totally agree.   This change doesn't change that at all.   That is specified in the dependency within each module and shouldn't be in the dependencyManagement also (which should just be for the version).

johanhaleby:
Oh thanks! Then I misunderstood.

PR #1078, opened 2018-10-11
Support for web test client

OlgaMaciaszek:
This adds support for Spring WebTestClient-based RestAssured Tests.

I have tried to maintain the original API and structure as much as possible. I have extracted some things to a new, shared module in order to avoid duplication. Finally, I have added the implementation with `WebTestClient` underneath; it mostly uses the existing `RestAssured` api. I have added the possibility to create router-function-based instances, requests with uri function paramters and the possibility to add consumer, however, I have not added the possibility to return a `WebTestClient` response spec for the users to run `WebTestClient`- style assertions, cause I have to close the body content stream in order to extract the byte[] body response content which is the base for all `RestAssured` assertions, which closes the stream and users would get errors if they try to execute `WebTestClient`-style body content assertions against that.


OlgaMaciaszek:
@johanhaleby I've filed the `WebTestClient` module PR as agreed. Please review.

PR #1081, opened 2018-10-20
Update Groovy version to 2.5.3

PascalSchumacher:
This requires an update of the Java version to 7 and the replacement of the gmaven plugin with the gmavenplus-plugin.

No sure if it is worth it.

### DIFF
--- a/json-path/pom.xml
+++ b/json-path/pom.xml
@@ -35,8 +35,8 @@
     <build>
         <plugins>
             <plugin>
-                <groupId>org.codehaus.gmaven</groupId>
-                <artifactId>gmaven-plugin</artifactId>
+                <groupId>org.codehaus.gmavenplus</groupId>
+                <artifactId>gmavenplus-plugin</artifactId>
             </plugin>
             <plugin>
                 <groupId>org.apache.felix</groupId>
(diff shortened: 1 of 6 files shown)


johanhaleby:
@PascalSchumacher Thanks for the PR! Unfortunately Groovy 2.5 requires Java 8 and REST Assured is supporting Java 6 so we can't update. I'll close this PR for now, but thank you anyway for your contribution.

PascalSchumacher:
Groovy 2.5 requires Java 7 not Java 8.

I not sure why anybody would still want to support Java 6 nowadays, but of course it is your call. 😃

johanhaleby:
> Groovy 2.5 requires Java 7 not Java 8.

Sorry I read the [release notes](http://groovy-lang.org/releasenotes/groovy-2.5.html) of Groovy 2.5 too quickly. It said that it requires Java 8 to *build* but only 7 to run so you're right.

> I not sure why anybody would still want to support Java 6 nowadays, but of course it is your call.

That might be the case. But if so there's lot's of stuff that can be made better, for example removal of a lot of code :) But sure, this should definitely be included in the release of REST Assured 4.0. And 5.0 should probably Java 11 only and then we can get rid of apache http client :)

PR #1090, opened 2018-11-12
allow unicode passwords in PreemptiveBasicAuthScheme

ludenus:


johanhaleby:
Hmm I think this is a bit too simplistic. From what I remember the standard says that "iso-8859-1" is to be assumed by default. I'd be happy to accept a PR that makes this configurable though, but in its current form this PR may lead to backward incompatibilities. 

Also sorry for the very late response.

ludenus:
hi @johanhaleby thanks for feedback. I am not going not insist on this PR and ready to close it because I am not really familiar with the standard. Just a few things for your information:
1.  Unicode passwords like "☠ЪZz12345" work well with swagger API client running against springboot services (not sure if springboot and swagger follow standards but both are quite common)
2.  I use rest-assured for testing but now I cannot test unicode passwords with rest-assured
3.  It seems ok to have iso-8859-1 by **default** but for rest-assured client there is no way to change password encoding, it is static final, so its not default, but hardcoded value and for the testing context feels a bit too restrictive

Issue #1093, opened 2018-11-17
Allow custom listeners on failure


Bisk1:
My initial problem was that I wanted to customize logging to be cleaner. We are sending a lot of headers and cookies in tests and it kind of clutters the output, so what I wanted was that: 
- full request and response to be always logged to DEBUG logger (handled by file appender) 
- simplified request and response (only URL + body) should go to INFO logger
This way I could have concise and short output, all the details landing in debug file which I check only when I need. 
However, this was not possible because the Request- and ResponseLoggingFilter are tightly coupled to 
1) specific format in RequestPrinter
2) options restricted to LogDetail (not sufficient for me)
3) single instance of PrintStream

So instead of adding more options to an already complicated filters I thought that I can write my own logging filter. The problem was, that I cannot use information whether validation failed or not just like the default filters do - because the code that does it refers default logRepository specifically. So for this I wanted to register my own listeners that would be invoked when there is a failure and provide relevant info as params (request spec, response spec and actual response object).

This could also solve the problem of people who want to see Response object after validation failed. For me most convenient would be listeners configured globally as there is no need to change logging logic on each request, but it would be also easy then to add method to spec builder, e.g. onFail(...)

I already have a tested branch with these changes and will soon submit PR. Let me know if this change makes sense.

gustahrodrigues:
Hi @Bisk1, thanks for your great work on that!

I'm trying to used it by looking at [this example](https://github.com/rest-assured/rest-assured/wiki/Usage#failure-config). My goal is to mask a request header with secrets on it, but I'm unable to do it, because ```reqSpec``` is always ```null```. Do you know why ? Could we fix it ?

It seems something related [to this line](https://github.com/rest-assured/rest-assured/pull/1094/files#diff-6611764b33ad2aa8b9cbbdc3c0beda10R502).

Thanks in advice,

gustahrodrigues:
I've just found a way to do it. If anyone else is looking for something similar, could use this:

```java
    ResponseValidationFailureListener validationFailureListener = (reqSpec, respSpec, resp) -> {
      final ResponseSpecificationImpl respSpecImpl = (ResponseSpecificationImpl) respSpec;
      final LogRepository logRepo = respSpecImpl.getLogRepository();

      final String requestLog = logRepo.getRequestLog();

     //process requestLog

      final ByteArrayOutputStream outStream = new ByteArrayOutputStream();
       try {
        outStream.write(requestLog.getBytes());
[...]

PR #1094, opened 2018-11-17
Improvement for #1093 - failure listeners 

Bisk1:
Failure listeners will allow API users:
- run custom code in response to validation failure
- access response objects after failure
- write more advanced filter with customized post-failure actions

I also refactored existing handling of 'logging on validation failure" to be one of these listeners, the only one to be configured by default.
I think the introduced listener could be used to further improve the code by removing workarounds related to logging e.g. in ValidatableResponseOptionsImpl.specification()

Detailed description for my use case in #1093

### DIFF
--- a/rest-assured/src/main/groovy/io/restassured/internal/LogRequestAndResponseOnFailListener.groovy
+++ b/rest-assured/src/main/groovy/io/restassured/internal/LogRequestAndResponseOnFailListener.groovy
@@ -1,0 +1,43 @@
+package io.restassured.internal
+
+import io.restassured.response.Response
+import io.restassured.specification.RequestSpecification
+import io.restassured.specification.ResponseSpecification
+import org.apache.commons.lang3.StringUtils
+import org.apache.commons.lang3.SystemUtils
+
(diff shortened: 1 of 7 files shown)


johanhaleby:
@Bisk1 Thanks a lot for the PR and I'm very sorry for the late response. I will look into this, hopefully soon!

Bisk1:
@johanhaleby OK, looking forward to your comments

johanhaleby:
@Bisk1 I'm so sorry for not getting back to you earlier. I've taken a look at the PR now and I think it looks very good and it's a good generalization of the "log if test failed use case" that also for better customizations.

johanhaleby:
I'm going to move `io.restassured.internal.ResponseValidationFailureListener` out of the `internal` package though since I don't think you mean it to internal right?

Bisk1:
@johanhaleby  Hey, I'm happy to see my PR merged and released.
Regarding your question - yes, `ResponseValidationFailureListener` should not be in `internal `package. Also sorry for late reply

Issue #813, opened 2017-02-19
Version-conflict in dependencies: httpclient:4.5.2 vs. httpmime:4.5.1


christopherfrieler:
I'm using the maven-enforcer-plugin and it fails, complaining about the dependencies of rest-assured. rest-assured depends on org.apache.httpcomponents:httpclient:4.5.2 and on org.apache.httpcomponents:httpmime:4.5.1, which again depends on org.apache.httpcomponents:httpclient, but version 4.5.1.

```
Dependency convergence error for org.apache.httpcomponents:httpclient:4.5.2 paths to dependency are:
+-my.group:artifact:1.0.0
  +-io.rest-assured:rest-assured:3.0.2
    +-org.apache.httpcomponents:httpclient:4.5.2
and
+-my.group:artifact:1.0.0
  +-io.rest-assured:rest-assured:3.0.2
    +-org.apache.httpcomponents:httpmime:4.5.1
      +-org.apache.httpcomponents:httpclient:4.5.1
```

Is there any reason not to use org.apache.httpcomponents:httpmime:4.5.2 or could you just update it?

As a workaround the version of httpmime can be raised using maven's dependency-management.
```
<dependencyManagement>
	<dependencies>
		<dependency>
			<groupId>org.apache.httpcomponents</groupId>
			<artifactId>httpmime</artifactId>
			<version>4.5.2</version>
		</dependency>
	</dependencies>
</dependencyManagement>
```

Additionally, not really a rest-assured-thing, but maybe someone knows, why maven complains about this anyway. I'd expected it to use org.apache.httpcomponents:httpclient:4.5.2, since the versions differ only in the third part.

johanhaleby:
I think this should work now thanks to PR mentioned above.

PR #1107, opened 2018-12-09
Keep version of httpmime in sync with httpclient

rherrmann:
`httpclient` is required in version 4.5.3, whereas `httpmime` is required in version 4.5.1. `httpmime` itself depends on `httpclient`, requiring the same version as its own version. This causes the maven enforcer plugin to complain about the dependency convergence when consuming rest-assured.

Use `${httpclient-version}` for all subcomponents of `org.apache.httpcomponents`.

See-also: #813


johanhaleby:
Thanks for the PR!

PR #1112, opened 2018-12-12
Create rest-assured-all to avoid split packages

### LINKED ISSUE #1111: Split packages when using xml path and json path
xml-path and json-path have the same root package, which causes errors when using both modules under Java 9 and higher.

TomaszGaweda:
Creating shaded jar allows people to use xml-path and json-path in one project

Closes #1111

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,7 @@
     <modules>
         <module>modules</module>
         <module>rest-assured</module>
+        <module>rest-assured-all</module>
         <module>examples</module>
         <module>json-path</module>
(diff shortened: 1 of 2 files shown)


johanhaleby:
Does this mean that users who experience this issue should depend on the `rest-assured-all` module instead their Maven/Gradle config?

TomaszGaweda:
@johanhaleby They should depend on `rest-assured-all`, not `xml-path` and `json-path`, then there will be no split packages - everything in one jar

PR #1123, opened 2019-01-26
Fix duplication of content type

geminiKim:
Hi :)
I found trivial bug for `contentType` and `headers` from MockHttpServletRequestBuilder.

This bug I think so trivial bug for rest-assured, but saw duplication of `content-type` in result docs when using with `spring-rest-dcos` and `asciidoctor`.

I added test code that case, if you could review this PR then I appreciate it.
AND.. I use rest-assured very usefully, Thank you for your effort. :D


johanhaleby:
Thanks, really appreciated.

PR #1126, opened 2019-02-04
versionrange added to osgi-dependency groovy

Rikske54:
Hello johanhaleby,

This is a fix for the osgi-dependency on groovy in rest-assured. The api of groovy has changed in release 2.5. I've added a versionrange so only groovy-versions between 2.4 and 2.5 are permitted, excluding 2.5 and onward.

Greetings,
Rikske54


johanhaleby:
Thanks a lot!

PR #1127, opened 2019-02-04
Fixed issue with overwriting FailureConfig on settings other configs

viclovsky:
Hi!
I set custom FailureConfig(https://github.com/rest-assured/rest-assured/pull/1094), but it was overwritten (created new one) on adding other configs.
See an example  below.
`...failureConfig(FailureConfig.failureConfig().failureListeners(new MyFailureListener()))).logConfig(getLogConfig())... `
There will be an empty FailureConfig config with default listener when we use similar chain of calles cause it was recreated.


viclovsky:
I just have found another issue
This one works: 
`given().config(RestAssuredConfig.config().failureConfig(failureConfig))
.get("http://jsonplaceholder.typicode.com/todos/1").then().statusCode(400);  `

But this one — doesn't : 
```
given().config(RestAssuredConfig.config().failureConfig(failureConfig))
.get("http://jsonplaceholder.typicode.com/todos/1").then().spec(new ResponseSpecBuilder().expectStatusCode(400).build());
```

johanhaleby:
Oh well spotted! Thanks a lot!

johanhaleby:
Regarding the second issue, is that resolved with this PR as well?

viclovsky:
> Regarding the second issue, is that resolved with this PR as well?

Second issue hasn't been fixed. I'll create another issue ticket

Issue #679, opened 2016-04-21
Add assert message to test


miwashiab:
given()
    .log().all()
.when()
    .get("/path")
.then()
    .statusCode(HttpStatus.SC_OK).assertMessage("Get /path must return statuscode OK!")
    .body("size()", greaterThan(0)).assertMessage("Get /path must return atleast one item!")

to get similar behaviour as
assertTrue("Must be true", condition);

johanhaleby:
Interesting idea. It'll probably not read like that though, more likely something like this:

```java
.. .body("size()", withDescription("Get /path must return atleast one item!"), greaterThan(0)). ..
```

Implementation note:

Expand "Argument interface" (`withArgs`) concept to be broader. Maybe "Argument interface" should extend from "Option" Interface or something.. Q: What about both `withArgs` and `withDescription`?

maciejsobala:
Any progress on this feature?

johanhaleby:
Sorry not yet :(

vvingolds-allpago:
Any news?

techbecca:
Know that this is a wanted feature! I am desperate to combine the customised error message with the logging on failed validation. The only solution I've come up with so far is to write my own assert class, which tries and catches the assertion errors from response validations and then adds the personalised message by throwing a new assertion errors.

MaciejWysocki:
I would like to have message code in
`then().statusCode(403)`
like this example:
`then().statusCode("Setup test user.", 403)`
or on different level like this:
`given().onFailMessage("Setup test user.")`

tsuhachev:
It would be also helpful in case I would like to extract a business failure message from the response body and include it into the assertion message, rather than search for it in the log.

Frechet:
Hi everyone. I would like to add another reason for this request.
Our  tests should be done in parallel, so we want to see a log with assert. 

Now it looks like this:
//log with another execution
java.lang.AssertionError: 1 expectation failed.
XML path items.size doesn't match.
Expected: is <3>
  Actual:  4
//log with another execution

So we should do this:
response.then()
.body("items.size"), describedAs("Expected 3. Fail. Request=" +  request  + ". Response="  + response, is(3)))
.body("items[0].name"), describedAs("Expected Kenobi. Fail. Request=" +  request  + ". Response=" +  response, is(3)))
... // about 10 -20 same lines

But a good solution for me should looks like this:
response.then()
.body("items.size"), is(3))
[...]

PR #1131, opened 2019-02-07
#679: Add on fail message builder for response specification

Frechet:
This is useful when you need to associate a mismatch description with additional user information. For example, in parallel execution.

### DIFF
--- a/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/ErrorMessageITest.java
+++ b/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/ErrorMessageITest.java
@@ -93,6 +93,32 @@ public class ErrorMessageITest extends WithJetty {
     }
 
     @Test public void
+    error_message_with_on_fail_message_look_ok_when_mixing_body_and_status_code_errors() {
+        exception.expect(AssertionError.class);
+        exception.expectMessage("3 expectations failed.\n" +
+                "Expected status code <201> but was <200>.\n" +
+                "\n" +
+                "JSON path lotto.lottoId doesn't match.\n" +
+                "Expected: a value less than <2>\n" +
+                "  Actual: 5\n" +
+                "\n" +
+                "JSON path lotto.winning-numbers doesn't match.\n" +
+                "Expected: a collection containing <21>\n" +
+                "  Actual: [2, 45, 34, 23, 7, 5, 3]\n" +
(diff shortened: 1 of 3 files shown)


johanhaleby:
Is it important that the failure message is a part of the AssertionError? 

I've made some changes where I'm using a [ResponseValidationFailureListener](http://static.javadoc.io/io.rest-assured/rest-assured/3.3.0/io/restassured/listener/ResponseValidationFailureListener.html) to print the error message instead. I like the idea that we're building on existing building blocks (`ResponseValidationFailureListener` in this case) instead of implementing completely new concepts. On my machine it now looks like this:

image

However since `describedAs` is a part of the AssertionError it probably makes more sense the way you did it. Just want to hear your thoughts.

Frechet:
@johanhaleby Thank you so much for your quick feedback!

> Is it important that the failure message is a part of the AssertionError?

I think, yes. I will try to explain why.
I looked at your proposed solution. Unfortunately, it cannot solve our problem. The fact is that in the ResponseValidationFailureListener we can write to the stream, but this will also be independent of the assertion written to the stream, as if we had written something like this:

```
log.info("test1");
response.then().statusCode(200)
--
[TestNG-tests-3] INFO test1
[TestNG-tests-2] INFO test3
// Here ResponseValidationFailureListener wrote something, but we cannot know it for this assert or another.
[TestNG-tests-4] INFO test2
...
java.lang.AssertionError: 1 expectation failed.
Expected status code <200> but was <404>.
[TestNG-tests-7] INFO test12
```

mih-kopylov:
@johanhaleby any chance to have this PR merged?

hasanalpzengin:
Important feature. What's status ? Do you need testers here ?

MRC81:
I guess many would be happy to have this feature! Any chances to see it merged?

cheparsky:
@johanhaleby maybe it is a good time to merge it? :) Do you need some testers or help?

cheparsky:
@Frechet There is a conflict, please resolve it, maybe after that @johanhaleby merge it

Frechet:
Hello everybody! Thanks for your feedback! I will try to resolve the conflict as soon as possible and ask for a merge.

Frechet:
@cheparsky , @johanhaleby resolved conflicts https://github.com/rest-assured/rest-assured/pull/1502

Issue #1129, opened 2019-02-05
Request-/Response is not logged when matcher throws AssertionError


danielFesenmeyer:
This issue is somehow related to #710, but not the same, because I'm using assertions "inside" RestAssured, not "outside".

RestAssured allowed to define assertions on Hamcrest matchers. Personally, I find hamcrest matchers quite verbose and try to avoid them, using standard JUnit assertions or AssertJ where possible. This works quite fine because I can do arbitrary assertions and throw arbitrary exceptions when implementing/extending my own Matcher (specifically the matches-method). But I found out that it does not work nicely together with the option _RestAssured.enableLoggingOfRequestAndResponseIfValidationFails()_. If inside the matches-implementation an AssertionError occurs, the request/response  **is not** logged. If inside the matches-implementation a RuntimeException occurs the request/response **is** logged. I would expect logging in both cases.
The issue can be reproduced with the following tests.

```
   private static final int WRONG_STATUS_CODE = 404;

    @BeforeClass
    public static void setup() {
        RestAssured.enableLoggingOfRequestAndResponseIfValidationFails();
    }

    @Test
    public void matcherReturningFalseLogsAsExpected() {
        given().when().get("https://www.google.com").then().statusCode(Matchers.is(WRONG_STATUS_CODE));
    }

    @Test
    public void matcherThrowingRuntimeExceptionLogsAsExpected() {
        given().when().get("https://www.google.com").then().statusCode(new TypeSafeMatcher<Integer>() {
            @Override
            protected boolean matchesSafely(final Integer actualStatusCode) {
                if (!Objects.equals(WRONG_STATUS_CODE, actualStatusCode)) {
                    throw new IllegalStateException("Unexpected status: " + actualStatusCode);
                }

                return true;
            }

            @Override
            public void describeTo(final Description description) {
                // not relevant here
            }
        });
    }

    @Test
    public void matcherThrowingAssertionErrorDoesNotLogAsExpected() {
        given().when().get("https://www.google.com").then().statusCode(new TypeSafeMatcher<Integer>() {
            @Override
            protected boolean matchesSafely(final Integer actualStatusCode) {
                assertEquals(WRONG_STATUS_CODE, (int) actualStatusCode);

                return true;
            }

            @Override
            public void describeTo(final Description description) {
                // not relevant here
            }
        });
    }
```

PR #1133, opened 2019-02-08
#1129: Supported for logging when thrown assertion errors from matcher internal

geminiKim:
Hi :)

Not logged when matcher throws AssertionError. #1129 
it had caught only `Exception` and i changed to catch `Throwable`

I added test code that case. if you could review this PR then I appreciate it.
Thank you for your effort. :D


geminiKim:
@johanhaleby I apologize for the mention. 
You may be busy, but I would appreciate your feedback on this.
I will be waiting that how you think about this PR. thank you. :D

johanhaleby:
Oh I've completely missed this. I'm so sorry. Looks like a good change to me.

Issue #1071, opened 2018-09-21
No support for MatcherConfig in RestAssuredMockMvcConfig


qlp:
`RestAssuredConfig` has support for a MatcherConfig (to configure the `ErrorDescriptionType`), but the `RestAssuredMockMvcConfig` has not.

As a result we cannot use the `ErrorDescriptionType.HAMCREST` in our mock mvc tests.

johanhaleby:
Feel free to supply a pull request. That would be really helpful.

qlp:
Please have a look at https://github.com/rest-assured/rest-assured/pull/1139 and let me know if I'm missing something.

PR #1139, opened 2019-02-23
Add support for MatcherConfig in RestAssuredMockMvcConfig (#1071)

qlp:
Still need to write unit test. Please let me know if more needs to be done (never contributed to rest-assured before).


johanhaleby:
Looks good to me :) Thanks a lot for contributing.

PR #1156, opened 2019-04-05
Supports using HttpStatus for the response status check.

geminiKim:
Hi :)

I improved the response status checking method using SpringFramework's HttpStatus, I wonder your opinion on this.
(It was changed same part of ValidatableWebTestClientResponse)

Thank you for your effort. :D


johanhaleby:
@geminiKim This is great, again a huge thanks for helping out. You're on fire :)

PR #1152, opened 2019-03-18
Supports using MediaType for Accept header value

geminiKim:
Hi :)
I'm wondering your opinion to configured using MediaType for Accept header value.
(This change is associated with #1142)

Thank you for your effort. :D

PR #1159, opened 2019-04-17
Supports using MediaType for Accept header in WebTestClient

geminiKim:
Hi :)
I added some code and tests for the WebTestClient supported accept value using a MediaType.
This Pull Request is similar #1152, but i've missed adding a code for the WebTestClient.

Thank you for your effort. :D


johanhaleby:
Again, great work :) Thank you!

PR #1164, opened 2019-05-02
Fixed typo in Javadoc of ValidatableResponseOptions.java

LorenzoBettini:
The Javadoc of T header(String headerName, ResponseAwareMatcher<R> expectedValueMatcher) shows an example that does not compile:

```
assertThat().header("Location", response -> response.endsWith("/x/") + response.path("id"));
```

in fact, `endsWith` is not a method of response, it's an Hamcrest matcher. This small patch fixes the example in the Javadoc. I propose the fix according to one of the tests of REST Assured itself, https://github.com/rest-assured/rest-assured/blob/master/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/HeaderITest.java#L220


johanhaleby:
Thank you!

PR #1184, opened 2019-06-20
Add unit tests for io.restassured.http.Header

petersear:
I've written some tests for the methods in `io.restassured.http.Header` with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes in a subsequent PR.

### DIFF
--- a/rest-assured/src/test/java/io/restassured/http/HeaderTest.java
+++ b/rest-assured/src/test/java/io/restassured/http/HeaderTest.java
@@ -1,0 +1,24 @@
+package io.restassured.http;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class HeaderTest {
+
+  @Test
+  public void header_has_same_name_as_expected() {
+    final Header header1 = new Header("foo", "bar");
+    final Header header2 = new Header("Foo", "baz");
+
+    assertThat(header2.hasSameNameAs(header1)).isTrue();
+  }
+
+  @Test
+  public void header_does_not_have_same_name_as_expected() {
+    final Header header1 = new Header("foo", "bar");
+    final Header header2 = new Header("bar", "baz");
+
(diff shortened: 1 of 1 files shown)


johanhaleby:
@petersear Thanks!

In my experience it's often not very useful to test equal/hashcode, tostring, getters/setters etc. `hasSameNameAs` is useful though. And also I don't really like when tests are named `testGetValue ` and `testGetName` etc. I'd much prefer to name tests so that they convey expected behavior and state and thus can be useful as (parts of) documentation and understanding intention. This talk is really a nice summary of what I'm talking about: https://2018.javazone.no/program/5c0fc222-ea41-48bc-a6a4-737ea171ffb7. I know I don't follow this nearly as much as I should in rest-assured project, mainly due to laziness on my part, but I really should. So although I'm very grateful for the offer to helpt write more tests I think I have to decline if they are structured they way you show in your PR (with the exception of `hasSameNameAs`, given that the test case is renamed). Does that make sense?

petersear:
@johanhaleby apologies for the delay in my response which is because I have been away for a couple of weeks. Thanks for your feedback on the tests. What you say makes perfect sense - I agree on both points and would be happy to modify the commit to reflect your suggestions. The intention here was to cover the whole class but I can see that it is not necessarily useful to cover some of those methods.
btw I have visited the webpage you quoted and will find time to listen to the talk in the next day or so, thanks.

petersear:
@johanhaleby I have removed the tests for all methods in the `io.restassured.http.Header` class apart from those for `hasSameNameAs()` as discussed. I separated the asserts into two separate tests which I have now named similarly to others in this project following your suggestions. Hopefully, I have understood your comments correctly. If you would like other changes to this PR, please let me know.
FYI I watched the talk and found it very interesting, thanks!

PR #1195, opened 2019-07-27
Fixed HeaderHelper, it did not regard content-type correctly when adding…

aschoerk:
… default characterset


aschoerk:
Hello,
during development of a module that should make it as easy to test jaxrs-resources as it is with Spring-MVC-Rest-controllers I encountered restassured. I took some tests from the Spring-MVC-restassured-module and adapted them to jaxrs testing. 
Please see this as a proposal. I would be glad to integrate the code I created in ioc-unit, to adapt to restassured, into the original, if you don't mind.

Naturally I am open to all suggestions.
Greetings Andreas

aschoerk:
Hello,
the pull request contains mostly new files (example test-files adapted from the spring-mvc-version) and almost no changes. It shows how you can create rest-assured-tests using the MockRequestDispatcher that is delivered by resteasy and i.m.O. is a quite nifty framework that you can use easily to test almost everything in one thread for what you otherwise needed to start a server. 

Perhaps anybody might have a quick look and comment on this?

PR #1198, opened 2019-08-04
#1197: do not set default port for non-localhost

### LINKED ISSUE #1197: RequestSpecBuilder sets 8080 port for non localhost uri
```
given().spec(new RequestSpecBuilder().build())
            .baseUri("http://nonlocal.com/api")
            .log().all()
            .filter(
                (requestSpec, responseSpec, ctx) -> new ResponseBuilder().setStatusCode(200).build()
                )
            .when().get();
```

Output is
```
Request method:	GET
Request URI:	http://nonlocal.com:8080/api
Proxy:			<none>
....
```

when expected output is:
```
Request method:	GET
Request URI:	http://nonlocal.com/api
Proxy:			<none>
```

dr29bart:
Fixes #1197 

replace port property reference with this.@port
this should skip the getter method which does some additional work.


johanhaleby:
Thanks for the PR!

PR #1201, opened 2019-08-09
Allow a user to specify their own AuthenticationScheme implementation

jamsesso:
In our use-case, we have an API-key based authentication scheme for internal service-to-service communication. Currently the only way for us to use these scheme in tests is to do something like this:

```
RestAssured.authentication = new ApiKeyAuthScheme("...");
given()...
```

This PR introduces a new syntax:

```
given()
  .auth().scheme(new ApiKeyAuthScheme("..."))...
```

### DIFF
--- a/rest-assured/src/main/groovy/io/restassured/internal/AuthenticationSpecificationImpl.groovy
+++ b/rest-assured/src/main/groovy/io/restassured/internal/AuthenticationSpecificationImpl.groovy
@@ -35,6 +35,13 @@ class AuthenticationSpecificationImpl implements AuthenticationSpecification {
     this.requestSpecification = requestSpecification
   }
 
+  def RequestSpecification scheme(AuthenticationScheme scheme) {
+    notNull scheme, "scheme"
+
+    requestSpecification.authenticationScheme = scheme;
+    return requestSpecification
+  }
+
   /**
    * Use http basic authentication.
    *
--- a/rest-assured/src/main/java/io/restassured/specification/AuthenticationSpecification.java
+++ b/rest-assured/src/main/java/io/restassured/specification/AuthenticationSpecification.java
@@ -16,6 +16,7 @@
 
 package io.restassured.specification;
 
+import io.restassured.authentication.AuthenticationScheme;
 import io.restassured.authentication.CertificateAuthSettings;
 import io.restassured.authentication.FormAuthConfig;
 import io.restassured.authentication.OAuthSignature;
@@ -24,6 +25,15 @@ import io.restassured.authentication.OAuthSignature;
  * Specify an authentication scheme to use when sending a request.
  */
 public interface AuthenticationSpecification {
+    /**
+     * Allows the user to provide their own {@link AuthenticationScheme} implementation for custom authentication
+     * mechanisms.
+     *
+     * @param scheme The authentication scheme implementation to use.
+     * @return The request specification.
+     */
+    RequestSpecification scheme(AuthenticationScheme scheme);
+
     /**
      * Use http basic authentication.
      *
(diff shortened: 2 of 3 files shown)


johanhaleby:
@jamsesso Hi! For some reason I've totally missed this PR. I'm sorry (to my defence I was on vacation 21 days ago).

I agree that it's strange that you can only set an `AuthenticationScheme` statically. However, if my memory serves me right, I don't think it was my intention that users would implement custom authentication strategies using this approach. My idea was that you would create [filters](https://github.com/rest-assured/rest-assured/wiki/Usage#filters) for this. The reason for exposing the `AuthenticationScheme` API was to allow global configuration of existing authentication schemes.

```
RestAssured.authentication = basic("username", "password");
```

However I can totally see why you'd like to create your own given the way the API is structured and that it's actually exposes the interface. The reason why I would still it's a bad idea to implement custom authentication this way is because the interface takes an instance of `HTTPBuilder`:

```groovy
void authenticate(HTTPBuilder httpBuilder)
```

HTTPBuilder is internal (well, it's located in the "internal" package) and it's not something I'd like people to depend on. I'm not sure how we should resolve this mess I've created :) Maybe I should move `AuthenticationScheme` to internal as well. 

However, it's definitely not obvious that you should use a filter to create custom authentication mechanisms given the way the API looks today. Maybe can somehow come up with a new interface that one can implement that we could pass to a method such as this:

```java
given().auth().custom(myCustomAuth). ..
```

The problem with this is that it'll probably be quite similar to that of a filter but arguably express the intent better.

WDYT?

jamsesso:
Not a problem re: vacation time. I just read up on filters, which I admittedly didn't know existed! What I like about the proposed interface is that it only allows a single auth mechanism to be defined. If I implement my authentication scheme as a filter, it may be confusing (by which I mean not immediately obvious) to the consumer of the API which auth scheme would be used if multiple are registered. For example:

```
.filter(new AuthSchemeFilterOne())
.filter(new AuthSchemeFilterTwo())
```

Of course you could use ordered filters, but in reality auth mechanisms are not typically additive - you're usually using one method or the other.

A concrete example would be my team: we have endpoints that exist only for service-to-service communication and are authenticated using an API key, while other endpoints exist for handling user requests and use a bearer token. The majority use-case for tests would be to use bearer token auth, so I would naturally define a global bearer auth scheme filter that I'd want to overwrite (not add to) for any requests that use API key auth.

To your hesitation of exposing HTTPBuilder, I agree - it's clearly an internal API. Perhaps we create something similar to `AuthenticationScheme` (`CustomAuthenticationScheme`? I'm having a hard time with naming at the moment) but closely resembles the interface of a filter. This way we can avoid the additive nature of filters and treat auth as a separate concept.

```
interface CustomAuthenicationScheme {
  void apply(FilterableRequestSpecification requestSpec, FilterContext ctx);
}
```

johanhaleby:
Regarding:

```java
interface CustomAuthenicationScheme {
  void apply(FilterableRequestSpecification requestSpec, FilterContext ctx);
}
```

maybe we should create two new interfaces instead of `FilterableRequestSpecification` and `FilterContext`? Not sure how difficult this will be to implement since you'd probably want to get access to the "authentication data" from filters and maybe even vice versa. Or do you think that all methods in `FilterableRequestSpecification` and `FilterContext` is applicable as they are from an auth point of view?

jamsesso:
I'm not sure, to be honest. What I was trying to illustrate was that the authentication scheme could have a similar interface to filters (and not rely on `HTTPBuilder`). The signature could be different, but we'd need the ability to attach query strings and headers at least. I'm not aware of any other auth mechanisms - but I'm sure some weird ones exist - perhaps injected into the request body?

PR #1215, opened 2019-09-19
changing HashMap to LinkedHashMap for deterministic iterations

contextshuffling:
When the ConfigurableJsonSlurper parses a json string, it uses a HashMap where iteration order is not guaranteed. When the resulting json object is prettified into a string again, the string can have elements in different order and thus, the test `io.restassured.path.json.JsonPathTest#jsonPathSupportsPrettyPrintingJson` may fail. This PR changes the HashMap to a LinkedHashMap, where iteration order is guaranteed. This PR also updates test assertions to match the deterministic order. 

Please let me know if you want to discuss the changes more.


johanhaleby:
This is great! Thanks a lot!

PR #1222, opened 2019-10-25
Blacklisting logged headers

sic2:
These changes will allow rest-assured clients to blacklist certain headers from being logged, regardless of whether the LogDetail is `ALL` or `HEADERS`.

The rationale of this change is that while headers are useful to look when debugging, they could also potentially contain secret or confidential data that one may wish to redact.

### Usage example:

```
new RequestSpecBuilder()
                .addHeader("x-api-key", "rest_assured_test")
                .addHeader("Authorization", "My funny secret")
                .log(ALL)
                .logBlacklists(new LogBlacklists()
                        .blacklistHeader("Authorization"))
                .build();
```

### Output example

```
Request method:	GET
Request URI:    http://localhost:8080/foo/bar
Proxy:          <none>
Request params: <none>
Query params:   <none>
Form params:    <none>
Path params:    x-api-key=rest_assured_test
Headers:        Accept=[ REDACTED ]
Cookies:        <none>
Multiparts:     <none>
Body:           <none>
```

### Further notes

These changes blacklist headers only, but the changes could be extended to other components of an HTTP request if required.

Also, this would be my first contribution to the project, so I hope this PR is fine and would be happy to take suggestions.

### DIFF
--- a/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/SpecificationBuilderITest.java
+++ b/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/SpecificationBuilderITest.java
@@ -399,7 +399,8 @@ public class SpecificationBuilderITest extends WithJetty {
                 "Form params:\t<none>%n" +
                 "Path params:\tfirstName=John%n" +
                 "\t\t\t\tlastName=Doe%n" +
-                "Headers:Cookies:\t\t<none>%n" +
+                "Headers:\t\tAccept=[ REDACTED ]%n" +
+                "Cookies:\t\t<none>%n" +
                 "Multiparts:\t\t<none>%n" +
                 "Body:\t\t\t<none>%n")));
     }
--- a/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/SpecificationBuilderITest.java
+++ b/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/SpecificationBuilderITest.java
@@ -19,6 +19,7 @@ package io.restassured.itest.java;
 import io.restassured.RestAssured;
 import io.restassured.builder.RequestSpecBuilder;
 import io.restassured.builder.ResponseSpecBuilder;
+import io.restassured.filter.log.LogBlacklists;
 import io.restassured.http.Cookies;
 import io.restassured.itest.java.support.WithJetty;
 import io.restassured.specification.RequestSpecification;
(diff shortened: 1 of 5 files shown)


sic2:
@johanhaleby - would you be able to look at this PR when you have some time? I would be very happy to get feedback from you.

johanhaleby:
@sic2 Sorry for not responding earlier, I have a crazy amount of stuff going on atm. 

Do I understand it correctly that the black list is empty by default? I.e. it doesn't break backward compatibility?

sic2:
@johanhaleby no problem. I was out of the country over the last few days too :)

The Blacklist entity is passed to the filter on all requests: https://github.com/rest-assured/rest-assured/pull/1222/files#diff-4da3dac70ae9b143ab367e220b30fb8eR112

but as you stated, it is initialised to be empty by default: https://github.com/rest-assured/rest-assured/pull/1222/files#diff-7e3b9a86fe2a2ac2d6840d76e4c97ff6R31


I have kept the existing test `supportsSettingLoggingWhenUsingRequestSpecBuilder` to ensure that we have backward compatibility on code not using this feature.

sic2:
@johanhaleby have you managed to have a look at this? cheers

johanhaleby:
@sic2 Sorry for the really late reply :(

If think that this is a nice contribution. However I have a question: Is there a special rationale behind placing the `logBlacklists` method in the `RequestSpecBuilder` class? I think it would make more sense to place `logBlacklists` in the `LogConfig` class. That way it would be more consistent with other configuration and we wouldn't need to extend the RequestSpecBuilder API. WDYT?

johanhaleby:
I've made some changes and merged to master. You now do like this instead:

```java
final RequestSpecification spec = new RequestSpecBuilder()
                .setConfig(newConfig()
                        .logConfig(logConfig().defaultStream(captor).blacklistHeader("Accept")))
                .and()
                .log(ALL)
                .build();

        given().
                spec(spec).
                pathParam("firstName", "John").
                pathParam("lastName", "Doe").
```

or:

```
given().
        config(config().logConfig(logConfig().blacklistHeader("Accept", "X-Something"))).
        log().all().
        pathParam("firstName", "John").
        pathParam("lastName", "Doe").
when().
        get("/{firstName}/{lastName}").
then().
        body("fullName", equalTo("John Doe"));
```

The blacklisted headers will also apply to response headers and they also work for the Spring specific modules (mockmvc webtest). WDYT?

sic2:
@johanhaleby thanks for the feedback. This was my first contribution and I was not familiar with the code base. 

I am glad you had a good look at it and corrected my mistake.

PR #1224, opened 2019-11-13
Fix typo in javadoc of log()

kthoms:


johanhaleby:
Thanks for noticing and PR :)

PR #1228, opened 2019-11-14
Jakarta EE API: switch from javax.xml.bind:jaxb-api to jakarta.xml.bi…

### LINKED ISSUE #1226: JAXB GAV changed: jakarta.xml.bind
See also:
 - https://wiki.eclipse.org/New_Maven_Coordinates

Would be awesome if rest-assured could change to using the new coordinates, as other projects are moving to change as well, but if you lag behind then we end up with non-conflicting GAVs , yet have conflicting classes.

I realize it's very annoying, don't shoot the messenger :) I just hope you could consider this.

Sanne:
…nd:jakarta.xml.bind-api

Fixes #1226 

N.B. The test `RestAssuredOSGiITest` is failing for me, but I had this problem before this patch as well so it seems unrelated. I don't really understand how that test is supposed to work so I hope it's ok I ignore that.


johanhaleby:
Thanks! The OSGi tests only works on Java 8 (see the [travis file](https://github.com/rest-assured/rest-assured/blob/master/.travis.yml)) unfortunately and they seem to pass so I guess it works :)

Thanks for your help!

Sanne:
many thanks to you!

Issue #882, opened 2017-07-06
IllegalArgumentException: No enum constant com.github.scribejava.core.model.Verb.CONNECT


vbachurin:
compile 'io.rest-assured:rest-assured:3.0.3'
    compile 'com.github.scribejava:scribejava-core:3.4.0'

I have added an OAuth2 authentication into the RestAssured configuration:

    RestAssured.authentication = oauth2(accessToken);

And one of my tests sends the DELETE, where I get the following exception:

    
    java.lang.IllegalArgumentException: No enum constant com.github.scribejava.core.model.Verb.CONNECT

	at java.lang.Enum.valueOf(Enum.java:238)
	at com.github.scribejava.core.model.Verb.valueOf(Verb.java:6)
	at io.restassured.internal.http.AuthConfig$OAuthSigner.process(AuthConfig.java:229)
	at org.apache.http.protocol.ImmutableHttpProcessor.process(ImmutableHttpProcessor.java:132)
	at org.apache.http.protocol.HttpRequestExecutor.preProcess(HttpRequestExecutor.java:166)
	at org.apache.http.impl.client.DefaultRequestDirector.createTunnelToTarget(DefaultRequestDirector.java:876)
	at org.apache.http.impl.client.DefaultRequestDirector.establishRoute(DefaultRequestDirector.java:796)
    ...
	at io.restassured.internal.RequestSpecificationImpl.delete(RequestSpecificationImpl.groovy:180)
	at io.restassured.internal.RequestSpecificationImpl.delete(RequestSpecificationImpl.groovy)
	at io.restassured.RestAssured.delete(RestAssured.java:838)
    ...

Indeed, there is no CONNECT in the corresponding enum:

https://github.com/scribejava/scribejava/blob/master/scribejava-core/src/main/java/com/github/scribejava/core/model/Verb.java

AllysonYuShiong:
@vbachurin Hi, I also facing the same error message. May I know did you get to solve this?

vbachurin:
@AllysonYuShiong I don't remember whether I was able to solve this in any way, sorry.

snpanigrahi9:
Getting same error in eclipse when calling twitter post api to publish twitt.
I am using rest assured java library. How to solve it?

shalama:
We are facing the same issue. In our case a POST but I think it does not matter.
We get this in combination with apache httpclient. If you use a proxy the http client is calling `HttpRequest connect = this.createConnectRequest(route, context);`. 

[...]

PR #1234, opened 2019-11-30
#882 - Ignore unsupported by scripbejava verbs

dlenroc:
Fix for #882, ignore verbs that are not supported by [scribejava](https://github.com/scribejava/scribejava/blob/master/scribejava-core/src/main/java/com/github/scribejava/core/model/Verb.java)


johanhaleby:
Nice find! Thanks for the contribution.

PR #1236, opened 2019-12-01
Make tests stable by changing test assertions

contextshuffling:
Tests in `io/restassured/module/mockmvc/ResponseLoggingTest.java` uses `jackson` to serialize objects into json and assert the serialized results with several hard-coded strings. However, the order of serialized json is not guaranteed so tests may fail if the order is different. So tests may fail or pass without any changes made to the source code and cannot serve as good regression tests.

This PR proposes to use `JSONAssert` and change test assertions to check JSON results in a safer way.


johanhaleby:
@contextshuffling Thanks a lot for the PR! Really appreciated.

PR #1254, opened 2020-01-10
Introduce spring-mock-mvc-kotlin-extensions

mhyeon-lee:
- kotlin-extensions module for rest-assured spring-mock-mvc


johanhaleby:
Oh cool stuff! I need to look into this a bit but it seems promising. Thanks a lot.

johanhaleby:
I think it's fine for now. Maybe we could do some fancy stuff to unify the two classes containing extension functions but I don't think it's worth the complexity. So great work, this will be a nice contribution for Kotlin users imho.

PR #1257, opened 2020-01-15
Introduce JSON-B object mapper type

aguibert:
Introduces a new mapper type for the Jakarta EE JSON Binding (JSON-B) specification. By default it will use Eclipse Yasson as the JSON-B implementation.

Testing is covered by `JsonPathBuiltinObjectDeserializationTest` which is parameterized to cover all values of `ObjectMapperType`.

Also adding a convenience configuration method to `RestAssured`, which simplifies user config for when they do want to override the default provider:
```java
// without new helper method
RestAssured.config = RestAssured.config.objectMapperConfig(ObjectMapperConfig.objectMapperConfig()
                                    .defaultObjectMapperType(ObjectMapperType.JSONB));

// with new helper method
RestAssured.objectMapper(ObjectMapperType.JSONB);
```

### DIFF
--- a/json-path/pom.xml
+++ b/json-path/pom.xml
@@ -100,8 +100,13 @@
             <optional>true</optional>
         </dependency>
         <dependency>
-            <groupId>javax.json</groupId>
-            <artifactId>javax.json-api</artifactId>
+            <groupId>jakarta.json</groupId>
+            <artifactId>jakarta.json-api</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse</groupId>
+            <artifactId>yasson</artifactId>
             <optional>true</optional>
         </dependency>
         <dependency>
--- a/json-path/src/main/groovy/io/restassured/internal/path/json/mapping/JsonObjectDeserializer.groovy
+++ b/json-path/src/main/groovy/io/restassured/internal/path/json/mapping/JsonObjectDeserializer.groovy
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 the original author or authors.
+ * Copyright 2019,2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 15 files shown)


johanhaleby:
@aguibert This is great! Thanks a lot for your contribution.

johanhaleby:
I'm a bit hesitant to add the `RestAssured.objectMapper(ObjectMapperType.JSONB)` though. Firstly I think it's a bit miss-leading since I've tried to differentiate between an `ObjectMapper` _instance_ and and `ObjectMapperType`. At least for me it would make more sense to call the method `RestAssured.objectMapperType(..)`. This is probably a minor but even `RestAssured.objectMapperType(..)` is somewhat miss-leading. I think a proper name would be `RestAssured.defaultObjectMapperType(..)` since you can override this configuration in tests later. This would make that point more clear. 

But I'm not sure how much people are actually setting a default object mapper (type) at all? I suspect that most people are happy with the default settings and for others I'm not sure if it legitimises a new API method for this shortcut. You could easily write a helper method yourself and delegate to

```java
RestAssured.config = RestAssured.config.objectMapperConfig(ObjectMapperConfig.objectMapperConfig()
                                    .defaultObjectMapperType(ObjectMapperType.JSONB));
```

Afaik a shortcut for configuring this has not been asked for before so maybe we can wait a bit see if more people request it. If so we can re-add the method. WDYT?

aguibert:
hi @johanhaleby, thanks for looking at my PR.

Regarding the helper method, as more ObjectMapper types get added, the default auto-detection mechanism will become less accurate. In my case, Jackson was already in my test dependencies somewhere and was getting picked as the default. Jackson has crept its way into a _lot_ of dependencies over the years. The reason I cared that my ObjectMapper was JSONB vs. Jackson is that my model classes were written to use JSONB and they failed to be [de]serialized if Jackson got selected.

For the current helper method approach, I could use that, but it took me a good 20-30 minutes of googling and inspecting the RestAssured code in order to figure out the right set of calls (the configuration API has a dangerous mix of static new-instance methods and static config-builder method).

Your points about method naming make sense. Would you be willing to compromise and have a `RestAssured.defaultObjectMapperType(..)` method?

aguibert:
@johanhaleby I was just looking over the commit where you removed the shortcut method, but you also removed an extra shortcut method that I did not add:
`public static void objectMapper(ObjectMapper objectMapper) {`
Was this intended? That could be a breaking change for your users

Issue #1255, opened 2020-01-14
Add support for SameSite in Cookie


anjo-swe:
Chrome 80 is about to be released (2/4) and it includes a major change to how cookies are handled.

> Cookies that are not properly labeled with **“SameSite=None”** and **“Secure”** will no longer be accessible across sites.

**References:**
* https://web.dev/samesite-cookies-explained/
* https://web.dev/samesite-cookie-recipes/

We are using Rest Assured to test our Services and it would be great if the `Cookie` class was improved with `SameSite` information. (Strict / Lax / None)

**Example Code:**
```
Cookie cookie =
    expect()
        .statusCode(204)
    .when()
        .get("/test/?write_cookie=foo")
        .detailedCookie("foo");

assertEquals("None", cookie.getSameSite());
```

johanhaleby:
Sounds like a nice contribution. Is it something you would consider helping out with?

anjo-swe:
@johanhaleby added a PR, let me know me what you think.

PR #1259, opened 2020-01-16
Cookie - Added support for SameSite Attribute

anjo-swe:
Fix for https://github.com/rest-assured/rest-assured/issues/1255


johanhaleby:
Thanks a lot!! I'll try to release a new version of REST Assured that includes this fix soon.

PR #1260, opened 2020-01-18
Allow CookieFilter to decide if cookie name has to be checked when storing cookies

bedla:
Hi,
I found that when I have have two cookies with same name eg. `JSESSIONID` with different paths eg. `/` and `/bar`, and called URI is like `/bar/xxx`, current implementation of `CookieFilter` does not send second cookie (in Chrome this kind of cookies are normally send to server).
I have created PR, so please check it if it is correct. I made this check configurable because of backward compatibility.
Thx,
Ivos


johanhaleby:
Thanks a lot :)

PR #1264, opened 2020-01-27
Multipart Support for PUT method

### LINKED ISSUE #1262: Multipart not supported with put() method.
Currently rest-assured doesn't support Multipart File with `PUT` method.  
Following Exception is thrown when we try to use `multipart` method with `put()` : `java.lang.IllegalArgumentException: Currently multi-part file data uploading only works for POST`

Version : 3.0.3

anki2189:
This PR fixes #1262

### DIFF
--- a/modules/spring-web-test-client/pom.xml
+++ b/modules/spring-web-test-client/pom.xml
@@ -82,6 +82,12 @@
             <artifactId>spring-webflux</artifactId>
             <version>${spring.version}</version>
         </dependency>
+        
+        <dependency>
+            <groupId>org.synchronoss.cloud</groupId>
+            <artifactId>nio-multipart-parser</artifactId>
+            <version>1.1.0</version>
+        </dependency>
 
         <!-- Test -->
         <dependency>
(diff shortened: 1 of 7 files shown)


anki2189:
@johanhaleby : Please review the changes :)

steve-todorov:
@johanhaleby ping :)

johanhaleby:
@steve-todorov Sorry for the very late response but thanks for the PR :)

I just have a question regarding the "nio-multipart-parser" dependency, why is this necessary? I'm always hesitant to include additional transitive dependencies.

modules/spring-web-test-client/pom.xml
```diff
@@ -82,6 +82,12 @@
             <artifactId>spring-webflux</artifactId>
             <version>${spring.version}</version>
         </dependency>
+        
+        <dependency>
+            <groupId>org.synchronoss.cloud</groupId>
+            <artifactId>nio-multipart-parser</artifactId>
```
johanhaleby: Why is this needed?

johanhaleby:
It looks like it works without this extra dependency so I'm going to merge it to master

steve-todorov:
Thanks for merging @johanhaleby! It looks like the `nio-multipart-parser` dependency is needed - the `master` is broken in Travis and on my machine as well. Adding the dependency back fixed it for me. Maybe @anki2189 can share more details if this needs to be scoped to only `test` or if it's actually needed? :)

modules/spring-web-test-client/pom.xml
anki2189: It was required for webflux to support multipart file upload.

steve-todorov:
By the way, do you plan to cut a release soon? We have a few tests which required this feature and had to disable them until this gets resolved.

PR #1267, opened 2020-02-05
 Fix for decoder config not being set on the RestAssuredResponse.

dawiking:
If the decoder config is not set on the response, lookup of default charset for a content-type defaults to using standard charset of the jvm. Standard charset of the jvm is partially controlled by system property `file.encoding`.


johanhaleby:
Thanks a lot for the PR!

Issue #1273, opened 2020-02-13
Json serialization fails when using the JsonbObjectMapper feature


wizzk42:
When I use the jsonb object mapper creating a io.restassured.path.json.mapper.factory. JsonbObjectMapperFactory the serialization fails because Jsonb does not know about writeObject.

Jsonb has conversion methods fromJson/toJson to deserialize/serialize Objects from/to JSON.

The error occurs here: https://github.com/rest-assured/rest-assured/blob/e8948ba6e6fda767033ce18eb65e9103735bdde7/rest-assured/src/main/groovy/io/restassured/internal/mapping/JohnzonMapper.groovy#L44

johanhaleby:
Thank you for helping out

PR #1274, opened 2020-02-14
fix jsonb serialization method in JsonbMapper #1273

wizzk42:
https://github.com/rest-assured/rest-assured/issues/1273#issue-564487405


johanhaleby:
Thank you!

PR #1277, opened 2020-02-29
Make tests more stable by using JSONAssert equals

testfixer:
Similar to the change you already merged for `ResponseLoggingTest.java`(#1236):

Tests in `io.restassured.module.webtestclient.LoggingIfValidationFailsTest.java` and `io.restassured.module.webtestclient.ResponseLoggingTest.java` use `jackson` to serialize objects into json and assert the serialized results with several hard-coded strings. However, the order of serialized json is not guaranteed so tests may fail if the order is different. So tests may fail or pass without any changes made to the source code.

This PR proposes to use JSONAssert and change test assertions to check JSON results in a safer way.


johanhaleby:
Thank you again :)

PR #1285, opened 2020-03-13
PrintableStream is the new interface for logging

dr29bart:
add new interface io.restassured.config.PrintableStream
non-compatible change: LogConfig#defaultStream returns PrintableStream now
overload existing PrintStream methods with the new type


johanhaleby:
Thanks for the PR. But what is the purpose of using a custom interface like `io.restassured.config.PrintableStream` instead of just using `PrintStream`?

Issue #1284, opened 2020-03-13
Support for defining CookieSpec in CookieFilter


anoopgarlapati:
We are using Restassured 4.2.0 in our testing framework.
In one of our flows, we perform redirects and one of the sites uses sets cookie in response using lowercase "set-cookie" header. As a result of this when we use CookieFilter, we are not able to capture these cookies as current implementation of CookieFilter uses DefaultSpec which is case sensitive. 
The current implementation of CookieFilter does not allow for defining the CookieSpec to use. However, if CookieSpec to be used is configurable then this would help capture our use-case as we can use a CookieSpec which is compatible with a different RFC than the default spec.

anoopgarlapati:
Closing the ticket as our application no longer has a need for this customization.

PR #1287, opened 2020-03-13
Allow support for specifying CookieSpec in CookieFilter

anoopgarlapati:
Hi
This PR is for Issue #1284
This PR adds support for specifying CookieSpec in CookieFilter.
My use case is captured in the issue linked above.
Please review the PR and let me know if anything can be improved.
Thanks !!
Anoop

### DIFF
--- a/rest-assured/src/main/java/io/restassured/filter/cookie/CookieFilter.java
+++ b/rest-assured/src/main/java/io/restassured/filter/cookie/CookieFilter.java
@@ -87,6 +87,30 @@ public class CookieFilter implements Filter {
         this.cookieStore = new BasicCookieStore();
     }
 
+    /**
+     * Create an instance of {@link CookieFilter} that allows specifying a CookieSpec.
+     *
+     * @param cookieSpec Specify whether or not to allow found two cookies with same name eg. JSESSIONID with different paths.
+     */
+    public CookieFilter(CookieSpec cookieSpec) {
+        this.cookieSpec = cookieSpec;
+        this.allowMultipleCookiesWithTheSameName = false;
+        this.cookieStore = new BasicCookieStore();
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


rest-assured/src/main/java/io/restassured/filter/cookie/CookieFilter.java
```diff
@@ -87,6 +87,30 @@ public CookieFilter(boolean allowMultipleCookiesWithTheSameName) {
         this.cookieStore = new BasicCookieStore();
     }
 
+    /**
+     * Create an instance of {@link CookieFilter} that allows specifying a CookieSpec.
+     *
+     * @param cookieSpec Specify whether or not to allow found two cookies with same name eg. JSESSIONID with different paths.
+     */
+    public CookieFilter(CookieSpec cookieSpec) {
+        this.cookieSpec = cookieSpec;
```
dr29bart: re-use existing constructors:
```
this(cookieSpec, false);
```

johanhaleby:
Hi, 

I've read your use case in the issue, but I'm hesitant to allow passing `CookieSpec` as is since it's not a class owned by REST Assured and it would expose HTTP Client in the API (which is something I'd like to avoid as much as possible). Maybe an alternative would be to add a `Function<io.restassured.http.Cookie, io.restassured.http.Cookie>` instead to transform the cookies before they're matched? In your case you could do something like this:

```java
new CookieFilter(cookie -> new Cookie(cookie.getName().toUppercase(), cookie.getValue());
```

WDYT?

anoopgarlapati:
@johanhaleby Sorry for the delay getting to this. 
Thanks for the review. I do like the alternative Function approach over introducing an external class. However, we no longer have a need for this feature request so I'm closing the PR and the linked issue.

PR #1303, opened 2020-03-30
Update to AssertJ 3.15.0

Riggs333:
For future updates of dependencies consider using tools like https://dependabot.com/


johanhaleby:
Thank you!

Issue #958, opened 2017-12-27
graphql testing


mrcasablr:
Wondering if there is graphql support planned.  This tool is incredible and as we move from rest to graphql, would love for graphql to be supported

johanhaleby:
There are no plans for that. But contributions are welcome :)

ksiv:
@mrcasablr , could you elaborate further. What is your issues with graphql?
graphql response is plain JSON and works out of the box all you need is to put your json request into requst body with about same amount of efforts as you do your params for REST request.

ansonliao:
@ksiv 
Could you provide a sample that rest-assured send graphql query to the backend, 

for exmple, 

URL: https://graphqlserver.com/graphql/v1
Headers: key = `apikey`, value = `zzzzzzzz`
Query String = 
```graphql
{
  user(filter: {entityId: "2164686"}) {
     id
  }
}
```

My code in `rest-assured` as below, and 

```java
public static void main(String[] args) {
        String url = "https://graphqlserver.com/graphql/v1;
        String headerKey = "apikey";
        String headerValue = "zzzzzzzz";

        String payload = "{\n" +
                "  user(filter: {entityId: \"2164686\"}) {\n" +
                "    id\n" 
                "  }\n" +
                "}";
        Map<String, String> b = new HashMap<>();
        b.put("query", payload);

        given()
                .contentType(TEXT)
                .header(headerKey, headerValue)
                .body(payload)
                .post(url)
                .then()
                .statusCode(200);
```

always got response code `500` from server, and this sample I migrated to Python is works well.

Please advise.

himgupta13:
Guys, is there any update on this? Do we support graphql API on rest assured?
I want to test Graphql API response through rest assured. 

```
URL: https://xyz/graphql/v1
Headers: content-type = application/graphql
Query String =
query { applyCoupon(fareKey: "DEL-BOM-AI887", providerId: 196, originalAmount: 2243, currency: "INR", isInternational: false, searchToken: "1q4h30lhdtpkdnhhdhphphqhh96ohdddptpduddvsvnwpk", numOfTravellers: 1, couponCode: "MONEY", otherSpend: 0) {coupon {data {couponCode message cashback {amount currency} instantDiscount {amount currency}} error {code message}} burn {data {burnAmount burnRule totalWalletBalance ixigoConvenienceFee} error {code message}}}}
```

On execution, getting below error:

```
[...]

PR #1328, opened 2020-05-18
Feature/graphql

cemnura:
Hello,

I was recently developing GraphQL documentation and quickstart for https://github.com/quarkusio 

The quickstart required an integration test. The default test library used in quarkus-quickstart (https://github.com/quarkusio/quarkus-quickstarts) projects are rest-assured.

So developed an integration test to call a GraphQL endpoint using rest-assured. Here is my PR including the integration test;

quarkusio/quarkus-quickstarts#548

I admit my solution was not the `prettiest`. I simply created a GraphQL request string and added it to the body and made a post to `"/graphql"` url.

Seeing that there where other users trying to make GraphQL integration test with rest assured  in issue #958.
This got me wondering if maybe there can be a way to include GraphQL request with rest assured.

I fiddled a possible way to expand rest-assured to be able to make GraphQL queries. The general idea was to create a GraphQL DSL to generate GraphQL queries to make a post request. Since the response of GraphQL queries will return JSON the `ResponseValidation` would not have to be expanded.

Since all GraphQL are post operations, I created a `graphql()` method and masked it to make a post to `/graphql` endpoint.

There are definitely missing parts and much more work to do. I wanted to create this PR to talk about the possibility of including GraphQL features in rest-assured.


cemnura:
Hi @johanhaleby  ,

what would you think about this?

johanhaleby:
@cemnura This is cool and it's something that I've seen requested by people as well. I'm a bit hesitant to introduce this to the "core" library though. If possible it would probably make more sense as a module. But since Java lacks extension functions it will probably be a bit more difficult, sort of like the spring-mvc support.

 One idea would be to implement it in Kotlin, maybe it could build upon the kotlin-extension-module. But maybe this will put off many users. Or maybe it we could use lombok to simulate extension functions?

WDYT?

cemnura:
Hello @johanhaleby, thanks for considering the graphql feature.

I would be interested to implement this feature in Kotlin. Are there any examples I can study?

Why would you think this feature would put off users? Would you think it would add complexity to the library making more confusing? Since rest-assured is a rest API testing library with the addition of this feature it would be more like a rest API testing library with an additional graphql API testing functionality.

I am not familiar with the usage of Lombok either then generating getter/setter? So, I did not completely understand your idea. What did you have in mind exactly?

ksiv:
@cemnura 
>  Since the response of GraphQL queries will return JSON the `ResponseValidation` would not have to be expanded.
According to the spec JSON in response it's not a part of specification but a temporal convenience.
Their main main objection is JSON is a limited format and they do not wanna be limited. 

 @johanhaleby
> I'm a bit hesitant to introduce this to the "core" library though.

Please give us more info about your concerns so we can start thinking in right direction. Don't loose this @cemnura  guy :-)

cemnura:
@johanhaleby 

What is your opinion on this feature? Would you consider adding graph-ql support? I would love to work on it with your guidance.

johanhaleby:
I'm so sorry for my lack of response. I think this PR is really cool, and I'd love to get it integrated. But preferably I'd like to have this as a module. But I'm not sure how it can be achieved in a nice way in Java :/ So I'm a bit unsure how to best proceed. 

One option would indeed be to convert to create a module that uses kotlin extension functions. These functions will generate static methods that you can use from Java. But the API won't look as nice from a Java perspective.. Another option would be to do something similar to the spring extension modules (see [here](https://github.com/rest-assured/rest-assured/tree/master/modules/spring-mock-mvc) for example). That would probably work, but it requires more work to implement it...  WDYT?

cemnura:
Hello @johanhaleby,

Thanks for your response. I will take a look at creating a module that uses kotlin extension functions and get back to you.

PR #1333, opened 2020-06-06
Upgrade activation 1.1.1 to 1.2.1 + use groupId jakarta.activation

ge0ffrey:
Removes dependencies javax.activation, com.sun.activation and org.apache.sling.javax.activation in favor of jakarta.activation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -318,22 +318,19 @@
             <dependency>
                 <groupId>jakarta.xml.bind</groupId>
                 <artifactId>jakarta.xml.bind-api</artifactId>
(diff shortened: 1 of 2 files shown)


johanhaleby:
Thank you!

ge0ffrey:
No worries.

In the mean time, I did find out in the meanwhile that for activation there are 2 jars:
- activation-api which only holds the API class
- activation which holds both the API and impl classes.
There is no activation-impl.

Currently you might not have any impl classes of activation in your classpath, but if that's true and as your unit tests are succeeded, you don't need them.

PR #1340, opened 2020-06-18
Rename Blacklist to Blocklist

jorsol:
Breaking change, but I leave it here anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,14 +5,14 @@
 [![Javadoc](https://javadoc-badge.appspot.com/io.rest-assured/rest-assured.svg)](http://www.javadoc.io/doc/io.rest-assured/rest-assured)
 
 
-Testing and validation of REST services in Java is harder than in dynamic languages 
-such as Ruby and Groovy. REST Assured brings the simplicity of using these 
+Testing and validation of REST services in Java is harder than in dynamic languages
+such as Ruby and Groovy. REST Assured brings the simplicity of using these
 languages into the Java domain.
 
 
-## News 
+## News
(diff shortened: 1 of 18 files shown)


johanhaleby:
@jorsol Thank you for the PR! 

If this wasn't a breaking change I would have pulled it in directly given the circumstances. But I haven't followed this debate closely enough, and if it's ok with you, maybe you could educate me a bit or point to a nice article of the etymology of this word? I've never personally associated blacklist with any racial related issues but I'm open for education :)

jorsol:
> If this wasn't a breaking change I would have pulled it in directly given the circumstances. But I haven't followed this debate closely enough, and if it's ok with you, maybe you could educate me a bit or point to a nice article of the etymology of this word? I've never personally associated blacklist with any racial related issues but I'm open for education :)

Well, I'm not the right person to ask... I also have never personally associated blacklist with any racial related issues either, the same happen with master/slave, but anyway humans are complex and there are biases that have connotations that could be misinterpreted as black*=bad and white*=good, and sometimes linguistic can also be weird, where you can use block to prevent from accessing or as building brick (just by the context), but again I'm not the right person to ask.

Having said that, feel free to close this PR if you think it's not worth the breaking change.

johanhaleby:
@jorsol I think it's nice that you brought it up. I have no intension on hurting anyone's feelings or use inappropriate naming conventions. I'm honestly a bit torn on what to do. I hope it's ok to wait a while until there's more consensus of naming conventions and a wide-spread unity regarding these things. So I'm going to close it for now, but I'm open to make changes in the future. Thanks for mentioning! :)

PR #1341, opened 2020-06-21
Fix content type name before charset recognition

jekanik:
Hi. I faced little bug in `HeaderHelper.class`. It is building default charset for content type in a such way:
```
if (encoderConfig.shouldAppendDefaultContentCharsetToContentTypeIfUndefined()) {
     contentType += "; charset=";
     if (encoderConfig.hasDefaultCharsetForContentType(contentType)) {
         contentType += encoderConfig.defaultCharsetForContentType(contentType);
     } else {
         contentType += encoderConfig.defaultContentCharset();
     }
}
return contentType;
```
So it means that content type is modified with "; charset=" part before passing this value to EncoderConfig#hasDefaultCharsetForContentType. And in my case with "application/json" I've got 
"application/json; charset=". Instead of getting charset as UTF-8 that is configured in EncoderConfig#DEFAULT_CHARSET_FOR_CONTENT_TYPE as default for json content type I've got ISO-8859-1 that is used as global default charset. 

Pls review and share your thoughts. Thanks.

### DIFF
--- a/modules/spring-mock-mvc/src/test/java/io/restassured/module/mockmvc/ContentTypeTest.java
+++ b/modules/spring-mock-mvc/src/test/java/io/restassured/module/mockmvc/ContentTypeTest.java
@@ -39,14 +39,14 @@ public class ContentTypeTest {
 
         RestAssuredMockMvc.given().
                 standaloneSetup(new GreetingController()).
-                contentType(ContentType.JSON).
+                contentType(ContentType.XML).
                 interceptor(requestBuilder -> contentType.set(extractContentType(requestBuilder))).
         when().
                 get("/greeting?name={name}", "Johan").
         then().
                statusCode(200);
 
-        assertThat(contentType.get()).isEqualTo("application/json;charset=" + RestAssuredMockMvc.config().getEncoderConfig().defaultContentCharset());
+        assertThat(contentType.get()).isEqualTo("application/xml;charset=" + RestAssuredMockMvc.config().getEncoderConfig().defaultContentCharset());
     }
 
     @Test public void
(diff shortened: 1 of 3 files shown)


modules/spring-mock-mvc/src/test/java/io/restassured/module/mockmvc/ContentTypeTest.java
```diff
@@ -101,10 +101,11 @@
     }
 
     @Test public void
-    doesnt_duplication_of_content_type_with_default_charset() {
+    doesnt_duplication_of_content_type() {
```
jekanik, line 104: I've removed duplicated test of adds_default_charset_to_content_type_by_default and added one more for testing my changes. Git is showing changeset in little bit strange way.

jekanik:
@johanhaleby pls review.

johanhaleby:
Thank you the PR and your patience, well spotted!

marcingrzejszczak:
In Spring Boot the current supported version of Rest Assured is 3.3.0 (https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-dependencies/build.gradle#L1380). Can we add this fix also to the 3.x version of Rest Assured?

leonard84:
Just FYI Spring removed the charset parameter completely spring-projects/spring-framework#22788 since https://tools.ietf.org/html/rfc7159#section-11 does not define any parameters for application/json.

marcingrzejszczak:
I see that 3.3.x is not maintained. Can this be cherry-picked to 4.2.x @johanhaleby ?

johanhaleby:
@marcingrzejszczak I tried backporting it to 3.x but it was really difficult and I have to give up. I'll try 4.2 next.

johanhaleby:
@marcingrzejszczak I actually think it worked! I'm really surprised myself :) 
image

johanhaleby:
I.e. version 4.2.1 should be available in maven central now

marcingrzejszczak:
Thanks @johanhaleby !

marcingrzejszczak:
Hey @wilkinsona can we bump Boot to use Rest Assured 4.2.1 ?

wilkinsona:
Yep: https://github.com/spring-projects/spring-boot/issues/22303.

PR #1361, opened 2020-07-18
According to RFC 7159, application/json should not have charset

michaldo:
RFC 7159:

> Added a note to the "IANA Considerations" clarifying the absence of a "charset" parameter for the application/json media type.

I know this effect can be achieved by flag "`shouldAppendDefaultContentCharsetToContentTypeIfUndefined`". My intention is change default behavior for application/json.

EncoderConfig: if application/json is removed from DEFAULT_CHARSET_FOR_CONTENT_TYPE, only text/json will remain. text/json is non-standard, undefined content type, so is't behavior is also undefined. I removed text/json and whole DEFAULT_CHARSET_FOR_CONTENT_TYPE. (text/json is abandoned concept, replaced by application/json)

UnicodeITest: I removed failing test. I debug the case and I found that response content type is set to "application/json; charset=ISO-8591-1". This content type is undefined, therefore expectation is undefined. I decide remove the test.

### DIFF
--- a/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/UnicodeITest.java
+++ b/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/UnicodeITest.java
@@ -60,15 +60,4 @@ public class UnicodeITest extends WithJetty {
                 body("value", equalTo("啊 ☆"));
     }
 
-    @Test public void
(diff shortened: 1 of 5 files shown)


johanhaleby:
While I don't like to change default behavior, I know that it's wrong to supply a charset in application/json. So with that in mind I think it would be good to accept this.

PR #1387, opened 2020-10-04
Describe non matching actual method with Hamcrest matcher format

bedla:
Hi,
I found that when I want to match json-path with float/double value Hamcrest matcher and this matcher fails, I am getting cryptic message that says that `3.14` is not equal to `3.14`.
This problem is when I put double constant to matcher, but value is parsed as float from Json payload. 
That do you think about this change?
Thx,
Ivos

original:
```
JSON path values.pi doesn't match.
Expected: <3.14>
  Actual: 3.14
``` 
after my change:
```
JSON path values.pi doesn't match.
Expected: <3.14>
  Actual: <3.14F>
```


johanhaleby:
This is awesome!! This has been a nuance for such a long time. I didn't know you could do something about it.

johanhaleby:
Thanks a lot :)

Issue #1392, opened 2020-10-18
RestAssured is not releasing connections when reusing HttpClient


dzieciou:
When resuing same HTTP client instance with `reuseHttpClientInstance()` configuration option I am getting the following error:
```
java.lang.IllegalStateException: Invalid use of BasicClientConnManager: connection still allocated.
Make sure to release the connection before allocating another one.
```
This does not happen when HTTP client is created for each subsequent request.

Here there are the steps to reproduce the both situations.

```java
import io.restassured.specification.RequestSpecification;
import org.mockserver.client.MockServerClient;
import org.testng.annotations.AfterMethod;
import org.testng.annotations.BeforeMethod;
import org.testng.annotations.Test;

import static io.restassured.RestAssured.config;
import static io.restassured.RestAssured.given;
import static io.restassured.config.HttpClientConfig.httpClientConfig;
import static org.mockserver.integration.ClientAndServer.startClientAndServer;
import static org.mockserver.model.HttpRequest.request;
import static org.mockserver.model.HttpResponse.response;

public class Bug {

    private MockServerClient mockServer;

    @BeforeMethod
    public void setup() {
        mockServer = startClientAndServer(8888);
        mockServer.when(request()).respond(response());
    }

    @Test
    public void buggy() {

        //@formatter:off
        RequestSpecification request = given()
                .baseUri("http://localhost")
                .port(8888)
                .config(config()
                        .httpClient(httpClientConfig()
                                .reuseHttpClientInstance()))
                .body("anything")
                .when();

        request.post("/");

        request.post("/");
        //@formatter:on
    }

    @Test
    public void worksFineWhenNotReusingHttpClientInstance() {

        //@formatter:off
        RequestSpecification request = given()
                .baseUri("http://localhost")
                .port(8888)
                .config(config()
                        .httpClient(httpClientConfig()))
                .body("anything")
                .when();

        request.post("/").then().extract().response();

        request.post("/");
        //@formatter:on
    }

    @AfterMethod
    public void after() {
        mockServer.stop();
    }

}
```

I am using REST-assured 4.3.1.

dr29bart:
Why do you want to use **_reuseHttpClientInstance()_** config?

dzieciou:
@dr29bart 

To avoid recreating HTTP client connection pool for each new test.

Keeping connection pool for a large test base seems to shorten test execution time.

dzieciou:
[...]

PR #1399, opened 2020-10-28
#1392 release used connections

dr29bart:


dr29bart:
.andReturn().body().asInputStream() will fail

PR #1416, opened 2020-11-11
Added a github action that runs mvn verify on each push

HanSolo:
On each push mvn verify will run on Ubuntu 18.04 using JDK 8, 11, 12, 13, 14, 15 and 16-ea

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+---
+name: Java CI
+
+on: [push]
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


johanhaleby:
Thanks a lot. I've been thinking about running the tests using github actions instead of travis since I will probably run out of credits. I'm a complete beginner of github actions though. Where/how can I see why the build results? :)

HanSolo:
Oh you can check in the Actions tab on the github repo page, there you will see a list of actions that have been executed and if you click on the action itself you can see the jobs on the left side. If you click on a job it will show you the output of the job. There you can check the build results.

PR #1424, opened 2020-12-02
Fix flaky test in WebTestClientSpecificationMergingTest

WenhaoSu:
The test `web_test_client_factory_is_not_overwritten_when_not_defined_in_specification()` in `modules/spring-web-test-client/src/test/java/io/restassured/module/webtestclient/WebTestClientSpecificationMergingTest.java` is flaky, which is caused by test-order dependency. `RestAssuredWebTestClient.reset()` should be called once at the beginning of `web_test_client_factory_is_not_overwritten_when_not_defined_in_specification()` to ensure the state is fine for later test runs. This pull request fixes this flakiness.

The flaky test was found by running iDFlakies (https://github.com/idflakies/iDFlakies) and the proposed fix method was found by running iFixFlakies (https://github.com/TestingResearchIllinois/iFixFlakies).


johanhaleby:
Thank you!

PR #1425, opened 2020-12-03
Ensure subclasses of ResponseOptions do not break Prettifier

samneirinck:
In Prettifier, when working with a subclass of `ResponseOptions` which is not a `RestAssuredResponseOptionsImpl`, a casting error occurs.

Seems like the `instanceof` check was missing some `()`.


johanhaleby:
Well spotted, thanks for the PR!

PR #1447, opened 2021-03-10
Strict Matcher requires same response String instance as argument

phaldan:
JsonPartMatcher from JsonUnit requires the same object instance as argument between `matches` and `describeMismatch` method.

https://github.com/lukas-krecan/JsonUnit/blob/4bb39ae2bf308b8ca6f1ac67725c08e07c0d7bcd/json-unit/src/main/java/net/javacrumbs/jsonunit/JsonMatchers.java#L199-L228


johanhaleby:
Could you describe a bit more what the problem is because I don't understand? Maybe it would be easier if you could provide a test case as well so that we can verify that this fixes something :)

PR #1458, opened 2021-04-11
Pull Request to solve a part of the Issue#1457

javpermar4:
This is a pull request so solve a part of the Issue#1457 where we find some grammatical errors and added a final to a constant because if it is public, it should be final. I have also deleted a part of a if condition where we use instance of where it checks if it is null or not so it wasn't needed to check if it was null in the if-condition and changed a isEqualTo to HasToString because it will check better what it is needed.


johanhaleby:
Thanks a lot :)

PR #1459, opened 2021-04-12
Added new content type multi-part

yusuf-tayman:
When working with multi-part, multi-part should be selected from Content type


johanhaleby:
Thanks for the PR, but isn't it a bit confusing since there can be many different subparts of a multipart? See for example https://docs.microsoft.com/en-us/previous-versions/office/developer/exchange-server-2010/aa493937(v=exchg.140)

yusuf-tayman:
> Thanks for the PR, but isn't it a bit confusing since there can be many different subparts of a multipart? See for example https://docs.microsoft.com/en-us/previous-versions/office/developer/exchange-server-2010/aa493937(v=exchg.140)

Yes, you're right the multi-part usage has become very common and I will make and post changes

PR #1473, opened 2021-05-21
Added new content type multi-part

yusuf-tayman:
The multi-part option was added because the Multi-part option could not be selected because it isn't in ContentType like _ContentType.JSON_.


johanhaleby:
Thank you!

Issue #1069, opened 2018-09-19
RestAssured fails to parse responses with cookies that has numbers out of signed decimal range 


Mjl33:
A request results in a response that has a cookie with a large number 
fails with `NumberFormatException`
e.g. for response with cookie `set-cookie: Max-Age=20000000`

I believe it's because cookies are parsed with `Integer.parseInt(s, 10)` which limits the range to signed integer decimal

danielblokus:
Correct me if I'm wrong but the RFC2616 says that max-age should be integer.

jsmyczek:
It doesn't say that it should be at most 31-bit integer, though.

mradomyski:
Issue still valid after 2 years. Maybe worth consideration?

johanhaleby:
@mradomyski Please provide a PR

PR #1494, opened 2021-07-24
Fix for #1069: fails to parse responses with cookies that has numbers out of signed decimal range

mradomyski:
Changed cookie maxAge from int to long as suggested by @Mjl33


johanhaleby:
Thank you :)

PR #1502, opened 2021-09-14
fix #679: Add on fail message builder for response specification

### LINKED ISSUE #679: Add assert message to test

Frechet:
Continuation for
https://github.com/rest-assured/rest-assured/pull/1131


johanhaleby:
Thank you!

Issue #1503, opened 2021-09-16
Default implementation of deprecated method detachRoot is incorrect


gboucherie:
https://github.com/rest-assured/rest-assured/blob/2389456a4f313b733c2c1c462f12686cd0e81f8f/rest-assured/src/main/java/io/restassured/response/ValidatableResponseOptions.java#L729

PR #1504, opened 2021-09-16
Fix default implementation of deprecated method detachRoot (#1503)

gboucherie:


johanhaleby:
Thanks for spotting this 🤦‍♂️😃

PR #1513, opened 2021-10-27
feat: a way to add more specific log details

lucas-nguyen-17:
- Now: RequestLoggingFilter only have one option LogDetail, there is no way to choose which parts will be logged. 
- Expected: log which request parts that specify.

We can do something like this:
```
RequestLoggingFilter.with(METHOD, HEADERS, BODY)
```


johanhaleby:
LGTM! Thanks for the PR, I'll include it in the next release.

PR #1522, opened 2021-12-08
ci: GitHub Actions JDK 8, 11, 17

sullis:


johanhaleby:
Unfortunately, rest assured doesn't compile with Java 17. I think it's an issue with gmaven.

PR #1562, opened 2022-03-21
Fixes OSGi itests and corrects groovy import range

jassuncao:
Downgrades pax-exam to 4.13.4 due to an issue with a slf4j import.

Corrects the Groovy import range.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <scm.branch>master</scm.branch>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <groovy.version>4.0.1</groovy.version>
-        <groovy.range>[3.0,4.0)</groovy.range>
+        <groovy.range>[3.0,5.0)</groovy.range>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -59,7 +59,7 @@
         <scm.branch>master</scm.branch>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <groovy.version>4.0.1</groovy.version>
-        <groovy.range>[3.0,4.0)</groovy.range>
+        <groovy.range>[3.0,5.0)</groovy.range>
```
ponziani, line 62: Wouldn't it be better to have range [4.0,5.0) ?

johanhaleby: @ponziani Indeed, my mistake! I changed the version after this commit.

johanhaleby:
This is great!! Thanks a lot for your help @jassuncao and @ponziani! This would have taken me days :)

PR #1567, opened 2022-03-24
Re-added groovy package to package imports

ponziani:
I'm afraid I removed some package imports that still exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,6 @@ REST Assured supports any HTTP method but has explicit support for *POST*, *GET*
 ## Documentation
 
 * [Getting started](https://github.com/rest-assured/rest-assured/wiki/GettingStarted)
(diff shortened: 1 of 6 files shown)


Yberion:
Hello,

- Yeah sorry wrong wording, I was thinking if that PR could fix my problem
- I'm using Gradle and there isn't any mention of `org.codehaus.groovy:groovy-xml` at all in my project (Spring boot, I can provide more info about packages if needed)

I should create an issue probably, my answer to this PR was because I was concerned if it could fix my problem.

ponziani:
* I'm not sure, but I don't think this PR will fix your issue. It shouldn't be too hard to test it though.
* I'm not familiar with Gradle, but you should be able to print the dependency tree somehow I guess. You need to know where this 3.x dependency comes from

johanhaleby:
Thank you!!

Issue #1493, opened 2021-07-24
default response specification never prints log though i have added log(LogDetail.ALL)


piyush-jhawar:
RequestSpecBuilder requestSpecBuilder = new RequestSpecBuilder().
        setBaseUri("https://api.postman.com").
        addHeader("X-Api-Key, "").
        setContentType(ContentType.JSON).
        log(LogDetail.ALL);
        RestAssured.requestSpecification = requestSpecBuilder.build();

        ResponseSpecBuilder responseSpecBuilder = new ResponseSpecBuilder().
        expectStatusCode(200).
        expectContentType(ContentType.JSON).
        log(LogDetail.ALL);
        RestAssured.responseSpecification = responseSpecBuilder.build();

        @Test
        public void validate_status_code() {
        get("/workspaces");
         }

Above test print request logs even though we don't pass any spec, but it doesn't print the response output.
 Question:
Is it necessary to add spec with then for response to print doesn't matter it is default responseSpecification or not.

My Understanding:
If it is default responseSpecification it should print log for response without passing the spec.
I mean this shouldn't be required then().spec(responseSpecification)

Code file attached for reference.
[codefile.txt](https://github.com/rest-assured/rest-assured/files/6871752/codefile.txt)

hetacz:
I have same issue, to get response I have to still add .then().log().all() :(

petur:
It looks like `SpecificationMerger` doesn't copy the `responseLogDetail` field, so setting it on `RestAssured.responseSpecification` has no effect. Would something like this fix it (untested)? https://github.com/rest-assured/rest-assured/pull/1599

petur:
Logging works as expected with #1599 applied.

PR #1599, opened 2022-05-27
Overwrite responseLogDetail when merging ResponseSpecificationImpl.

petur:
Calls to `setLogDetail` on `RestAssured.responseSpecification` had no effect because the field wasn't copied when creating the request.

fixes: #1493


johanhaleby:
Thank you!

PR #1605, opened 2022-06-21
Update RequestSpecification.java

Kerry-G:
Changed pathParameter to pathParam in RequestSpecification.java


johanhaleby:
Thank you :)

PR #1625, opened 2022-09-24
chore: allow contentType() to accept Spring MediaType

hantsy:
Similar to `accept` which allows developers to use Spring Mediatype as Content Type, add `contentType(...)` to accept Spring MediaType.


johanhaleby:
Thank you! :)

Issue #1638, opened 2022-10-15
Method Response.asPrettyString() destroys binary responses


BChabowski:
Behaviour of `Response.asPrettyString()` seems to be different than `Response.asString()`. Regular `asString` method just returns response as String, wheras `asPrettyString` also saves string version of response as new body and - in process - destroys binary data. Please consider following code:

```java
public static void main(String[] args) throws IOException {
        //download dummy pdf
        RestAssured.baseURI = "https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf";

        //this will work fine
        Response firstResponse = given()
                .contentType(ContentType.BINARY)
                .when()
                .get();

        firstResponse.asString();
        parsePdf(firstResponse.asInputStream());

        //this will result in exception since asPrettyString converts response body
        Response secondResponse = given()
                .contentType(ContentType.BINARY)
                .when()
                .get();

        secondResponse.asPrettyString();
        parsePdf(secondResponse.asInputStream());
    }

    private static void parsePdf(InputStream inputStream) throws IOException {
        PDDocument document = PDDocument.load(inputStream);
        if (!document.isEncrypted()) {
            PDFTextStripper stripper = new PDFTextStripper();
            String text = stripper.getText(document);
            System.out.println("Successfully parsed pdf from input stream:" + text);
        }
        document.close();
    }
```

In `RestAssuredResponseOptionsGroovyImpl.groovy` string is explicitely saved as new body:
```groovy
  String asPrettyString(ResponseOptions responseOptions, ResponseBody responseBody) {
    def body = new Prettifier().getPrettifiedBodyIfPossible(responseOptions, responseBody)
    content = body
    body
  }
  ```
  Why is that? Is this intentional to convert the body only in this method? There is no info about that in documentation, and the bahaviour is unexpected. 
  The problem arose in our project, where we log every response before using it. It seems easy to repoduce, so it's likely we are not the first or last to encounter it.
  
  Please advice whether it is a bug or something we need to live with.

**Comment by lucas-nguyen-17** [comment]
Pls calm down. It's open source, you can expect it MUST support all people needs.

Back to the problem, I don't see any different from both methods. I tested your code as well, and no exception.
- `getPrettifiedBodyIfPossible` will make string pretty if it has content-type which is one of JSON, XML, HTML.
- If not (no defined content-type or not define content-type at all), it will fall back to `.asString()`

You can take a look the code here: class Prettifier

```
[...]

PR #1639, opened 2022-10-18
Removed content assignment from asPrettyString()

BChabowski:
Content assignment in `asPrettyString()` was converting whole response body into string, which was problematic when `asPrettyString()` was used first, and then binary content of response was needed. 
What's more, this behaviour wasn't in line with the behaviour of `asString()` method.

It is explained more in depth in linked issue, please take a look at it: https://github.com/rest-assured/rest-assured/issues/1638


johanhaleby:
Thanks a lot!

Issue #1631, opened 2022-10-04
Be able to send a keystore with relaxedHTTPSValidation


mathieu-amblard:
I would like to implement Service Integration Tests using rest-assured.
The main idea is to execute some requests to validate that the service has been correctly deployed.
This service is behind an intranet proxy configured with HTTPS, therefore I would like to send a keystore and trust all certificates.

If I configured the SSL from like this :
```java
given()
    .config(newConfig().sslConfig(
            new SSLConfig()
                    .keyStore(myKeystore, myKeystorePass)
                    .relaxedHTTPSValidation()
    ))
...
```
I get the following exception :
```text
400 Bad Request : No required SSL certificate was sent
```
I was expected the keystore to be sent but as you can see it was not the case.

As a workaround, I have configured the SSL using the SSLSocketFactory :
```java
...
KeyManagerFactory kmf =
        KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
kmf.init(myKeystore, myKeystorePass.toCharArray());
KeyManager[] kms = kmf.getKeyManagers();
...
SSLContext sslContext = SSLContext.getInstance();
sslContext.init(kms, null, null);
SSLSocketFactory factory = new SSLSocketFactory(sslContext,
        SSLSocketFactory.ALLOW_ALL_HOSTNAME_VERIFIER);
...
given()
    .config(newConfig().sslConfig(
            new SSLConfig()
                    .sslSocketFactory(factory)
    ))
```
What is annoyed me with the workaround is that the SSLSocketFactory is deprecated (already addressed with #1610) and especially that the `relaxedHTTPSValidation` does almost the same (except the keystore configuration).

Would it be possible to have a ` .relaxedHTTPSValidation()` method containing parameters for loading a keystore or that uses a keystore configured with the method `keystore(...)` ? 

If you agree to add a such functionality, I can make you a proposal in a Pull Request if needed.

Thank you in advance

mathieu-amblard:
PR Created

PR #1640, opened 2022-10-19
Enable the use of relaxedHTTPSValidation with two-way ssl

mathieu-amblard:
Please refer to : https://github.com/rest-assured/rest-assured/issues/1631


johanhaleby:
Thank you for the PR! Looks like a great contribution. Are you aware of any backward-incompatibilities?

mathieu-amblard:
Thanks for your reply,
As far as I know, there should not be any backward-incompatibilities as it is a new feature.
All previous tests are still working without any changes, so I guess these changes do not break anything.

johanhaleby:
Thanks for your comment and for the PR :)

PR #1652, opened 2022-11-17
Spring Framework 6 is never resolved

marcingrzejszczak:
with this change Spring Framework 6 version is never actually found cause we're looking at a non-existant class

with Spring Framework 6 GA out we can safely use another class that was introduced in 6.0 - org.springframework.aot.AotDetector


johanhaleby:
Thanks for the PR! Will make a new release today.

PR #1667, opened 2022-12-21
Accept cookies that have an expiration date

### LINKED ISSUE #1430: Cookies with expiry date is not handled correctly in CookieFilter
Cookies which contains an expiry date that is not adhering to the 'expired date format' specified at 

> http://web.archive.org/web/20020803110822/http://wp.netscape.com/newsref/std/cookie_spec.html

 is not getting applied in subsequent requests. 

The cookie expiry date format that is now recommended by **IETF is - Expires=Fri, 07 Feb 2020 12:46:00 GMT** (https://tools.ietf.org/html/rfc6265). HttpClient though has not updated to this format yet and expects a format - **Expires=Fri, 07-Feb-2020 HH:MM:SS GMT**. Note the difference in dashes between IETF and Netscape formats. 

> HttpClient further recommends to use custom configuration of Cookie Policy (https://www.mail-archive.com/dev@hc.apache.org/msg16624.html).

Though we can set the httpClient cookie config, CookieFilter does not use the standards from CookieSpec. If we modify the below code in CookieFilter.java 
```
public CookieFilter(boolean allowMultipleCookiesWithTheSameName) {
        this.allowMultipleCookiesWithTheSameName = allowMultipleCookiesWithTheSameName;
        this.cookieSpec = new DefaultCookieSpec(); //This uses the default dateFormat from netscape
        this.cookieStore = new BasicCookieStore();
    }
```

to 

```
public CookieFilter(boolean allowMultipleCookiesWithTheSameName) {
        this.allowMultipleCookiesWithTheSameName = allowMultipleCookiesWithTheSameName;
        this.cookieSpec = new DefaultCookieSpec(new String[]{"EEE, dd MMM yyyy HH:mm:ss z"},false);
        this.cookieStore = new BasicCookieStore();
    }
```
the problem might get resolved. I can raise a PR if this approach sounds good.

FroMage:
Fixes #1430

### DIFF
--- a/rest-assured/src/main/java/io/restassured/filter/cookie/CookieFilter.java
+++ b/rest-assured/src/main/java/io/restassured/filter/cookie/CookieFilter.java
@@ -23,12 +23,14 @@ import io.restassured.response.Response;
 import io.restassured.specification.FilterableRequestSpecification;
 import io.restassured.specification.FilterableResponseSpecification;
 import org.apache.http.Header;
+import org.apache.http.client.CookieStore;
 import org.apache.http.cookie.Cookie;
 import org.apache.http.cookie.CookieOrigin;
 import org.apache.http.cookie.CookieSpec;
 import org.apache.http.cookie.MalformedCookieException;
 import org.apache.http.impl.client.BasicCookieStore;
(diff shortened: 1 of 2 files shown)


johanhaleby:
Thank you!

Issue #1211, opened 2019-09-11
Scala support library for RestAssured is not cross compiled


xeppaka:
Shouldn't scala support library be cross compiled? We use scala 2.12.9 and as the latest rest assured (v4.1.1) scala-support is compiled with scala 2.13 we can't use it in the project.

johanhaleby:
@xeppaka Yes it probably should! Can you do this easily with maven? Care to provide a PR if you know how to do it?

xeppaka:
> @xeppaka Yes it probably should! Can you do this easily with maven? Care to provide a PR if you know how to do it?

Don't know how it's done with maven... I'll try to find out

PR #1670, opened 2023-01-18
Cross compile to Scala 2.11, 2.12 & 2.13

atais:
fixing #1211 
Sorry for copying of the files, but with Maven it will be really difficult to achieve this task.

Maybe splitting scala_support at some point would be more beneficial? 
And maintaining it separately? It's not like it is changing a lot.

The CI does not work properly for anything else but JDK8, but in master I see only JDK8 being used...
So I am a bit confused how to work with that, since it fails with newer JDK on parts I did not touch.


atais:
I have actually just now realized, that the scala support... it quite an empty project.

What's it purpose? It causes more confusion then actually it has to offer.

johanhaleby:
Thanks a ton for your effort, but yeah, it's really just a very simple extension function so it might not be worth the effort of maintaining it and introducing all this complexity. For people that want the extension function, maybe it's good enough to just document the code instead and let people copy it to their projects. WDYT?

PR #1678, opened 2023-02-16
Fix convertCookies to copy missing attributes

### LINKED ISSUE #1462: MockMvcRequestSenderImpl convertCookies function is not copying the httpOnly, sameSite and expires from the servletCookies into the cookieBuilder
The convertCookies function in io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl is not copying the httpOnly, sameSite and expires attributes from the servletCookies to the cookieBuilder resulting in an incorrect result.

As the servlet spec's javax.servlet.http.Cookie does have support for the sameSite and expires attributes, a instance of check to org.springframework.mock.web.MockCookie will need to be made to copy the sameSite and expires attributes from there.

justin-tay:
Fixes #1462

Copies `isHttpOnly` from the servlet cookie and copies `getExpires` and `getSameSite` which can be found on `org.springframework.mock.web.MockCookie`.

See https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/main/java/org/springframework/mock/web/MockCookie.java

Fix convertCookies to copy missing attributes

### DIFF
--- a/modules/spring-mock-mvc/src/main/java/io/restassured/module/mockmvc/internal/MockMvcRequestSenderImpl.java
+++ b/modules/spring-mock-mvc/src/main/java/io/restassured/module/mockmvc/internal/MockMvcRequestSenderImpl.java
@@ -65,6 +65,7 @@ import java.lang.reflect.Constructor;
 import java.net.URI;
 import java.net.URL;
 import java.security.Principal;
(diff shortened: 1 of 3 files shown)


johanhaleby:
Thanks a lot!

PR #1683, opened 2023-03-20
fix: wrong path in PathSupport when param has http scheme

### LINKED ISSUE #1682: Unexpected logic in PathSupport.getPath()
While developing an OAuth authentication server, I discovered unexpected behavior when the query parameter in the path contains an http scheme. 

As shown in the following example, when the targetUri is `/oauth/authorize?client_id=clientId&redirect_uri=https://sample.com&response_type=code&login_type=token`, the code at line 61 searches for the index of "://" and treats it as the host's scheme. Consequently, when attempting to find the index of "/", the path is not found, resulting in a bug where the path is "/".

image

To address this issue, either the path should be encoded before being evaluated if the content type is application/x-www-form-urlencoded, or the logic of the method needs to be modified.

My current rest assured version is 5.2.1. Please let me know if this has been fixed in a different version.

fucct:
fix #1682

### DIFF
--- a/rest-assured/src/main/groovy/io/restassured/internal/support/PathSupport.groovy
+++ b/rest-assured/src/main/groovy/io/restassured/internal/support/PathSupport.groovy
@@ -58,9 +58,9 @@ class PathSupport {
       return targetUri
     }
 
-    def indexOfScheme = targetUri.indexOf("://");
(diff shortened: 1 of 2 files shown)


KIMSIYOUNG:
this is what i need

johanhaleby:
Thanks a lot for the PR!

PR #1722, opened 2023-08-30
Make header blacklisting case-insensitive

bekh6ex:
Motivation: 

Many load balancers override header values, and developers grow a habit to ignore case sensitivity of the headers. 
In practice header names are treated as case-insensitive and it can potentially lead to leaking credentials in logs. 
Speaking of standards and HTTP RFCs: some of them state header fields are case insensitive, some of them state otherwise. 

Also, having some sensible secure defaults can be beneficial.

This can prevent many security incidents of type "credentials in build pipeline logs".


johanhaleby:
Thank you for the PR! It's great that they are case-insensitive now, great find. 

However, I'm not sure that I want to add the headers you provided as sensible by default for backward compatibility reasons. I'll merge the PR, but I'll make some changes to this.

bekh6ex:
👍

Issue #1548, opened 2022-02-08
onFailMessage not available via posted example


elmuerte:
The 4.5.0 release notes provides this example:

```java
when().
      get().
then().
      onFailMessage("Some specific message").
      statusCode(200);
```

But that does not work. The `onFailMessage` is only part of the `ResponseSpecification`, and not the `RequestSpecification` or `ValidatableResponseOptions`.

This is possible:

```java
expect().
      onFailMessage("Some specific message").
      statusCode(200).
when().
      get();
```

jrnewton:
@johanhaleby - I'm happy to contribute to fix this one but would like to know if the doc is wrong or the code is wrong.

JuliaResh:
@jrnewton Would it be possible to make `onFailMessage` available both in `ResponseSpecification` and `ValidatableResponseOptions`?

dtitar:
Hi @jrnewton, any updates on this task?

jrnewton:
@dtitar - sorry but not using rest-assured right now so don't have the cycles.  My [original question](https://github.com/rest-assured/rest-assured/issues/1548#issuecomment-1089574832) still stands as well:  is the doc wrong or the code wrong?

PR #1724, opened 2023-09-03
add onFailMessage into validatable response

groov1kk:
Fix for the #1548

Added method *onFailMessage* into *ValidatableResponseOptions* to add possibility overriding an error message from *then* section:

```java
when().
       get("/lotto").
then().
       onFailMessage("An additional information to find the cause of the error").
       body("lotto.lottoId", lessThan(2));
```

UPD: Here is a [link](https://github.com/rest-assured/rest-assured/wiki/Usage#log-if-validation-fails) to the documentation with mistake which was discussed in the issue topic.


johanhaleby:
I thought that this already worked!! But it seems like it only worked for `expect` and not for `then`. Great find, and thanks for PR!

PR #1726, opened 2023-09-05
Correction, removed the content = string equalization

andrey67895:
There are situations when the built-in print() function is needed - it is convenient. But due to the presence of a content change to string inside it, format-sensitive data may break. For example, the method can receive as String (in case of an error) so is File, but if there is a file, converting the original content to String led to an error when debagging.


johanhaleby:
If I remember it correctly, `content` is used for caching purposes, if you try to read the response again. This will be quite a large breaking change if I bring it in. And I don't see any tests.

PR #1731, opened 2023-09-24
Introduce spring-web-test-client-kotlin-extensions

TheDevLuffy:
* kotlin-extensions module for rest-assured spring-web-test-client


johanhaleby:
Thank you, this is great!

PR #1747, opened 2023-11-06
Fix 1746: JsonPath: wrong result and exception when jsonPath partly evaluates to string/boolean/number

### LINKED ISSUE #1746: JsonPath: wrong result and exception when jsonPath partly evaluates to string/boolean/number
Given json:
```json
   { "item": {"subItem": "value"}}
```
When I evaluate the path:
```
"item.subItem.xx" 
```
then I get exception:
```
The parameter "xx" was used but not defined. Define parameters using the JsonPath.params(...) function
caused by:
  groovy.lang.MissingPropertyException: No such property: xx for class: java.lang.String
```
**What happens here:**
Groovy evaluates the `"item.subItem"` to `"value"` and then tries to invoke the `"value".xx` on String object, which cause the `MissingPropertyException`. RestAssurred catches this exception, then incorrectly interprets this as missing parameter and propose user the incorrect fix: _Define parameters using the JsonPath.params(...) function._

This happens when part of the json-path (`item.subItem....`) is evaluated to non-null and non-object value (i.e. string/boolean/number). 
To fix that we need to analyze the message of MissingPropertyException [here](https://github.com/rest-assured/rest-assured/blob/master/json-path/src/main/groovy/io/restassured/internal/path/json/JSONAssertion.groovy#L47)

For the case when we have missed parameter ( `new JsonPath("{}").get("$==someParam")` the exception will be:
```
groovy.lang.MissingPropertyException: No such property: someParam for class: Script1
``` 
So we need to return `IllegalArgumentException`  only when the message matches to string above (`No such property: .* for class: Script1`). This should catch only properties defined at top-level, i.e. properties which should be defined by user.
In another case need to return the `null`

kool79:
Closes #1746

### DIFF
--- a/json-path/src/main/groovy/io/restassured/internal/path/json/JSONAssertion.groovy
+++ b/json-path/src/main/groovy/io/restassured/internal/path/json/JSONAssertion.groovy
@@ -45,9 +45,14 @@ class JSONAssertion implements Assertion {
         }
         result = eval(root, object, expr)
       } catch (MissingPropertyException e) {
-        // This means that a param was used that was not defined
-        String error = String.format("The parameter \"%s\" was used but not defined. Define parameters using the JsonPath.params(...) function", e.property)
-        throw new IllegalArgumentException(error, e)
+        def message = e.getMessage();
+        // detect missed property on script-level. This should be defined by user as param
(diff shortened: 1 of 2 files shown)


johanhaleby:
Thanks a lot again!!

PR #1770, opened 2024-02-05
Create scala-extensions module for Scala 3

carlosedp:
This PR adds a scala-extensions module for a Scala 3 DSL similar to the existing Kotlin DSL. This extension is important since rest-assured is heavily used in Quarkus tests for endpoints and goes along the [Quickstart](https://github.com/carlosedp/scala3-quarkus-quickstart) for Scala 3/Quarkus I've been working on.

There are some points of improvement that could be done before merging (I might need some help):

- [ ] The method `Then` always returns a `ValidatableResponse` so an `Extract` could be chained. It would be great if it could return `Unit` in case there is no `Extract` so there won't be the need to declare the test `def` return type as `Unit` (to be picked-up by JUnit). I added a `ThenAssert` convenience method that uses `Then` but returns `Unit` for simplified usage.
- [ ] The `Extract` method needs the type to be defined in the `val/var` it's been assigned to. Don't know if there's a way around this.
- [x] ~~Extracting an `Int` from json requires specifying the type as a Java `Integer` instead of a Scala `Int`. This generates the error `java.lang.Exception: Unsupported type`.~~   **Solved.**

Let me know any additional requirements to have this in the main Rest-assured tree
This is based on the great work by @gavinray97

There are a couple of Scala code in-tree that I'd like to hear some feedback on how to tackle. Should we keep this module in Scala 3 or should I rewrite in Scala 2.13?

- [ ] Deprecate scala-support module?
- [ ] Update examples/scala-example
- [ ] Update examples/scala-mock-mvc-example
- [ ] Update examples/scalatra-example
- [ ] Update examples/scalatra-webapp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ target
 .settings
 .springBeans
 .idea
+.metals
+.vscode
+.bloop
 classes/
 lib/
 rest-assured-all/dependency-reduced-pom.xml
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,22 @@
+version = "3.7.15"
+runner.dialect = scala3
+maxColumn = 120
+align.preset = most
+align.multiline = false
+continuationIndent.defnSite = 2
+assumeStandardLibraryStripMargin = true
+docstrings.style = Asterisk
+docstrings.wrapMaxColumn = 80
+lineEndings = preserve
+includeCurlyBraceInSelectChains = false
+danglingParentheses.preset = true
+optIn.annotationNewlines = true
+newlines.alwaysBeforeMultilineDef = false
+rewrite.rules = [RedundantBraces]
+
+rewrite.redundantBraces.generalExpressions = false
+rewriteTokens = {
+  "⇒": "=>"
+  "→": "->"
+  "←": "<-"
+}
--- a/modules/pom.xml
+++ b/modules/pom.xml
@@ -15,7 +15,8 @@
   ~ limitations under the License.
   -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <parent>
         <artifactId>rest-assured-parent</artifactId>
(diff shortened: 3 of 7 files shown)


carlosedp:
Regarding item 3:

```scala
// This only works if userID is an Integer (Java Integer). It does not work with Scala Int
        val userID: Integer = Given()
            .When(req =>
                req.contentType("application/json")
                req.body(
                    """|{"name": "John Doe",
                       | "email": "blabla@blabla.com",
                       | "hashedPassword": "12345678"
                       |}""".stripMargin
                )
                req.post("/users")
            ).Then(res =>
                res.statusCode(200)
                res.body("name", is("John Doe"))
                res.body("email", is("blabla@blabla.com"))
            ).Extract(_.path("id"))
        // This works fine when userID is a Scala Int
        val userID: Int = `given`()
            .when()
            .contentType("application/json")
            .body(
                """|{"name": "John Doe",
                   | "email": "blabla@blabla.com",
                   | "hashedPassword": "12345678"
                   |}""".stripMargin
            )
            .post("/users")
            .`then`()
            .statusCode(200)
            .body("name", is("John Doe"))
            .extract().path("id")
```

I wonder why it doesn't convert to a Scala `Int`... I get:

```
[ERROR] org.acme.persistence.magnum.UserResourceTest.getUsers -- Time elapsed: 1.554 s <<< ERROR!
java.lang.Exception: Unsupported type
	at helper.RestAssuredHelper$package$.Extract(RestAssuredHelper.scala:171)
	at org.acme.persistence.magnum.UserResourceTest.createUser(UserResourceTest.scala:36)
	at org.acme.persistence.magnum.UserResourceTest.getUsers(UserResourceTest.scala:57)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at io.quarkus.test.junit.QuarkusTestExtension.runExtensionMethod(QuarkusTestExtension.java:1013)
	at io.quarkus.test.junit.QuarkusTestExtension.interceptTestMethod(QuarkusTestExtension.java:827)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

Line 36 pointed by the error is the .Extract line.

carlosedp:
Latest push cleans-up some redundancy and fixes a couple of errors.
Tests and functionality are pretty much aligned with Kotlin module.

johanhaleby:
Wow this is awesome! Thanks for the PR, I'll bring it in.

When it comes to deprecations of the "scala-support" module, I'm not really sure what to say because I don't use Scala myself nowadays. I like that you're using scala 3, but I'm not sure if most people are using Scala 3 or 2.13 nowdays.

carlosedp:
Cool, thanks for the merge Johan!
I'll look into the examples if then might get an update but I believe the tests are pretty self-explanatory.

GavinRay97:
Nice, thanks for doing this!

I forgot I wrote that in 2021 -- many years later and still using Quarkus, so always good to have an improved experience there =)

Issue #537, opened 2015-07-21
Unknown status code throws when using ResponseLogFilter


johanhaleby:
_From [eskato...@gmail.com](https://code.google.com/u/115671622875546989676/) on January 10, 2015 12:07:23_

What steps will reproduce the problem? 1. Use rest-assured@2.4.0
2. Call RestAssured.enableLoggingOfRequestAndResponseIfValidationFails()
3. Invoke HTTP resource that return an unknown code, like 601 What is the expected output? What do you see instead? java.lang.IllegalArgumentException: Status code must be greater than 100 and less than 600, was 601.
        at com.jayway.restassured.builder.ResponseBuilder.build(ResponseBuilder.java:175)

Without logging enabled things works well.

Enabling logging should not fail tests that pass without it.

From my understanding, HTTP RFCs do not forbid usage of status codes >599, they are simply unspecified.

_Original issue: http://code.google.com/p/rest-assured/issues/detail?id=382_

johanhaleby:
_From [johan.ha...@gmail.com](https://code.google.com/u/105676376875942041029/) on January 16, 2015 11:06:28_

You're right. Thanks for reporting.

**Status:** Accepted

macmon:
Hi,

I created a pull request [1773](https://github.com/rest-assured/rest-assured/pull/1773) where I removed the check from the Builder class. We really need this fix to check our application because we're using "911" as application specific error code.

PR #1773, opened 2024-02-28
issues/537: Remove check for response HTTP status code.

macmon:
Fix [Issue 537](https://github.com/rest-assured/rest-assured/issues/537) by removing the response status code check from the response logging filter code.


johanhaleby:
Very true! Thanks for the PR.

Issue #1751, opened 2023-11-23
RequestSpecification.queryParam does not respect null values


maxfd3v:
RequestSpecification.queryParam ("myParam", null)

Will be passed to the target API as a "" (Blank String)

One has to test the incoming parameter and skip adding a queryParam to make it work

DeMol-EE:
Not sure what @maxfd3v expects to happen, but I for one would like it if the query param was simply skipped when the passed value is null. I realise that it’s possible to wrap adding the query param in an if with a null check, but it looks verbose, especially when used in a parametrized junit 5 test:

compare
```java
        RequestSpecification req = RestAssured.given();
        if (foo != null) {
            req.queryParam("foo", foo);
        }
        if (bar != null) {
            req.queryParam("bar", bar);
        }
        Response res = req.get();
```

to
```java
        RequestSpecification req = RestAssured.given()
            .queryParam("foo", foo)
            .queryParam("bar", bar);
        Response res = req.get();
```

I feel like it would be easy to draft a PR for this, but the problem is that it would be breaking, unless we add a new method, like `optionalQueryParam`?

beng9re:
I opened a PR (#1776) to implement the functionality described in that part.

injae-kim:
`optionalQueryParam()` will be so useful to many users include me :)

PR #1776, opened 2024-03-18
Implement optionalQueryParam Following Discussion in #1751

beng9re:
This PR introduces the `optionalQueryParam` method to the `MockMvcRequestSpecification` class, as discussed in issue #1751. 

## Implementation Details
The `optionalQueryParam` method enhances request building by conditionally adding query parameters:

- **When the QueryParam value is null:** No query parameter is added to the request.
- **When the QueryParam value is not null:** The query parameter is added with the specified name and value.

## Discussion Reference
This implementation follows the specifications and considerations discussed in issue #1751, aiming to address the need for optional query parameters in our request handling logic.

## Testing
Tests have been added to verify that the `optionalQueryParam` method behaves as expected, ensuring that parameters are correctly added only when their values are not null.

### DIFF
--- a/modules/spring-mock-mvc/src/main/java/io/restassured/module/mockmvc/internal/MockMvcRequestSpecificationImpl.java
+++ b/modules/spring-mock-mvc/src/main/java/io/restassured/module/mockmvc/internal/MockMvcRequestSpecificationImpl.java
@@ -300,6 +300,15 @@ public class MockMvcRequestSpecificationImpl implements MockMvcRequestSpecificat
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


johanhaleby:
@beng9re Thanks again for a nice PR. However, I'm a little hesitant to merge this because I suspect (but I could be wrong) that it's a rather slim use case and it would add significantly more methods to the API (we should add it for form parameters as well etc). I understand that's a bit cumbersome to deal with this in Java, but in Kotlin (which I use most of the time) it's easy to write your own extension function. Also, this PR only address it for mock-mvc, not the "core", kotlin or web-test API so it would require more work to make it consistent. So before I put effort into this, I'd like to hear your thoughts.

Issue #1728, opened 2023-09-16
IllegalArgumentException: Only Strings allowed in path parameters


hantsy:
I tried to upgrade my Spring Boot 3.1/Java 17 example to Spring Boot 3.2-M2 and Java 21, the restassured-mock-mvc test failed.

The example project: https://github.com/hantsy/spring6-sandbox/tree/master/boot

The failed tests: https://github.com/hantsy/spring6-sandbox/blob/master/boot/src/test/java/com/example/demo/web/PostControllerTestWithRestAssuredMockMvc.java

```bash
[ERROR] Tests run: 9, Failures: 0, Errors: 6, Skipped: 1, Time elapsed: 10.97 s <<< FAILURE! -- in com.example.demo.web.PostControllerTestWithRestAssuredMockMvc
[ERROR] com.example.demo.web.PostControllerTestWithRestAssuredMockMvc.testGetPostById -- Time elapsed: 0.591 s <<< ERROR!
java.lang.IllegalArgumentException: Only Strings allowed in path parameters.
        at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.lambda$validateUnnamedPathParams$1(MockMvcRequestSenderImpl.java:328)
        at java.base/java.util.Optional.ifPresent(Optional.java:178)
        at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.validateUnnamedPathParams(MockMvcRequestSenderImpl.java:327)
        at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.applyPathParams(MockMvcRequestSenderImpl.java:416)
        at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.sendRequest(MockMvcRequestSenderImpl.java:347)
        at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.get(MockMvcRequestSenderImpl.java:653)
        at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.get(MockMvcRequestSenderImpl.java:105)
        at com.example.demo.web.PostControllerTestWithRestAssuredMockMvc.testGetPostById(PostControllerTestWithRestAssuredMockMvc.java:109)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)

[...]

PR #1782, opened 2024-04-23
Remove validation pathParams

taeyeon-Kim:
resolved: #1728

params will be expanded to string in the code below.
https://github.com/spring-projects/spring-framework/blob/adc7f731707d954c8991d8aacfc9a6f9f8c2fc34/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java#L156

https://github.com/spring-projects/spring-framework/blob/adc7f731707d954c8991d8aacfc9a6f9f8c2fc34/spring-web/src/main/java/org/springframework/web/util/UriComponents.java#L303-L305


johanhaleby:
@taeyeon-Kim Thanks for the PR!

PR #1796, opened 2024-07-05
Bump actions. Test with JDK 8/11/17/21.

carlosedp:


carlosedp:
I think there might be some incompatibilities with newer JDKs.

PR #1745, opened 2023-11-05
Migrate from GMaven to GMavenPlus

keeganwitt:
Splitting the migration to GMavenPlus in #1742 into a separate PR.

PR #1798, opened 2024-07-09
Migrate to GMavenplus

keeganwitt:


keeganwitt:
I meant to open this against my fork, but GitHub changed the default behavior when opening a PR so it opened here. It's OK, it can sit as draft for now.

appiepollo14:
@keeganwitt Great pr. one minor detail, can you describe why the two tests are to be ignored? Ps. I think a reviewer needs to be requested for this one to be moved.

keeganwitt:
> @keeganwitt Great pr. one minor detail, can you describe why the two tests are to be ignored? Ps. I think a reviewer needs to be requested for this one to be moved.

I think I commented in another PR about that. They were failing, but I didn't know enough about RestAssured to fix them.

See #1745 for related conversations.

PR #1801, opened 2024-08-05
fix: GZIP becomes gzıp for Turkish environments

rdmrtn:
"GZIP".toLowerCase() -> gzıp in Turkish.
This causes gz1p to be sent until httpclient 4.5.9 and gz?p since 4.5.10 due to httpcore 4.4.12 upgrade. This causes error for some header-sensitive servers.

Thx @frknpg


johanhaleby:
Nice find, thank you!

PR #1809, opened 2024-11-03
Use Long in maxAge(Matcher<? super Integer> maxAgeMatcher)

julianladisch:
Trying to use DetailedCookieMatcher with maxAge(is(greaterThan(1L)) doesn't compile because of the Integer - Long mismatch.

Calling DetailedCookieMatcher's maxAge(is(greaterThan(1)) compiles but fails during the test:

```
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
```

Fix:

Change

```
    public DetailedCookieMatcher maxAge(Matcher<? super Integer> maxAgeMatcher) {
```
to
```
    public DetailedCookieMatcher maxAge(Matcher<? super Long> maxAgeMatcher) {
```

Then maxAge(is(greaterThan(1L)) compiles and succeeds during the test.

This fix doesn't affect maxAge(is(1)) and maxAge(is(1L)). Both compile but the former keeps failing during tests and test latter keeps succeeding during tests.

### DIFF
--- a/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/CookieITest.java
+++ b/examples/rest-assured-itest-java/src/test/java/io/restassured/itest/java/CookieITest.java
@@ -67,7 +67,10 @@ public class CookieITest extends WithJetty {
         given()
                 .get("/multiCookie")
                 .then()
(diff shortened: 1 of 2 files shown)


johanhaleby:
Thanks for the PR! Nice find :)

PR #1820, opened 2025-05-08
Caching the length/size in a variable to avoid repeated calls in loops

LuisOsv:
Caching the length/size in a variable to avoid repeated calls in loops, which can slightly improve performance and readability.

Motivation and Context
While reading High Performance with Java ([link](https://learning.oreilly.com/library/view/high-performance-with/9781835469736/B21942_03.xhtml#_idParaDest-63)), I came across a useful performance tip: it's recommended to cache the size of a collection before entering a loop, instead of calling .size() in the loop condition.

Instead of:

```
for (int i = 0; i < list.size(); i++) {
    // logic
}
```
Prefer:

```
int size = list.size();
for (int i = 0; i < size; i++) {
    // logic
}
```

### DIFF
--- a/modules/spring-commons/src/main/java/io/restassured/module/spring/commons/HeaderHelper.java
+++ b/modules/spring-commons/src/main/java/io/restassured/module/spring/commons/HeaderHelper.java
@@ -124,7 +124,8 @@ public class HeaderHelper {
             String headerName = header.getName();
             if (headerConfig.shouldOverwriteHeaderWithName(headerName)) {
                 int index = -1;
-                for (int i = 0; i < filteredList.size(); i++) {
+                int filteredListSize = filteredList.size();
+                for (int i = 0; i < filteredListSize; i++) {
                     Header filteredHeader = filteredList.get(i);
                     if (filteredHeader.hasSameNameAs(header)) {
                         index = i;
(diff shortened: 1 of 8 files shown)


johanhaleby:
Are you sure that the JIT compiler won't fix this automatically?

LuisOsv:
Hi @johanhaleby  I did a quick test using JMH benchmark and there is a slightly improvement. 
[cached_size_benchmark_results_v7.txt](https://github.com/user-attachments/files/20210561/cached_size_benchmark_results_v7.txt)

| Benchmark                                       | Mode | Cnt |   Score   |   Error   | Units |
|------------------------------------------------|------|-----|-----------|-----------|-------|
| CachedSizeBenchmark.iterateWithForLoopCached   | avgt |  10 | 270.411   | ± 41.816  | ns/op |
| CachedSizeBenchmark.iterateWithForLoopNotCached| avgt |  10 | 281.851   | ± 61.816  | ns/op |

You can find on https://github.com/LuisOsv/benchTests repo

johanhaleby:
@LuisOsv Unfortunatley I had to revert because it failed a test: LoggingIfValidationFailsTest.logging_of_both_request_with_uri_function_and_response_validation_works_when_test_fails

PR #1825, opened 2025-06-28
Remove the cast of handling Uri template in Spring WebTestClient

### LINKED ISSUE #1824: UriTemplate handling error in RestAssuredWebTestClient
I used the following stack in my example project, encountered an issue: 

* Java 21
* Spring Boot 3.5.3
* WebFlux
* `spring-web-test-client` 5.5.5

The following does not process correctly. 

```java
.get("/posts/{id}", id) // here id is `Long` type
```

It throws exceptions. 

```bash
java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.lang.String (java.lang.Long and java.lang.String are in module java.base of loader 'bootstrap')

```

But the Uri template building worked well in pure `RestAssured` or `RestAssuredMockMvc`.

hantsy:
Compared to the code fragment of Spring MockMvc integration module, I remove the cast operation. Please review and suggest the changes. 

fix #1824

### DIFF
--- a/modules/spring-web-test-client/src/main/java/io/restassured/module/webtestclient/internal/WebTestClientRequestSenderImpl.java
+++ b/modules/spring-web-test-client/src/main/java/io/restassured/module/webtestclient/internal/WebTestClientRequestSenderImpl.java
@@ -593,7 +593,7 @@ public class WebTestClientRequestSenderImpl implements WebTestClientRequestSende
 		do {
 			final String paramName = pathParamMatcher.group(1);
 			getPathParamValueFunction.apply(paramName).ifPresent(paramValue ->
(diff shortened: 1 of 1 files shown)


hantsy:
In the source codes, it looks like the URI template in path is a little complex. 

In spring mockmvc and webtestclient integration modules, it can utilize Spring UriTemplate or DefaultUriTemplateHandler to process the values in URI template path safely.

johanhaleby:
@hantsy Hmm are you sure that this works? I'm not sure if there are tests that checks this functionality, but just removing URL encoding doesn't seem right.

hantsy:
Just compare the codes with webmvc and remove it. If it is handled by Spring uriTemplate handler, I think it will handle the URI encode automatically.

PR #1826, opened 2025-06-30
Attempt to upgrade Yasson to 3.0.3 with minimal changes.

### LINKED ISSUE #1651: Support JSON-B 3.0 (Jakarta EE 10)
Jakarta EE 10 being out, we now have the [JSBON-B 3.0](https://github.com/jakartaee/jsonb-api/releases/tag/3.0.0) specification and [Yasson 3.0](https://projects.eclipse.org/projects/ee4j.yasson/reviews/3.0.0-release-review) implementation being released with the new [package name](https://github.com/jakartaee/jsonb-api/tree/master/api/src/main/java/jakarta/json/bind) `jakarta.json.bind`.

So far, RestAssured 5.2.0 only supports Yasson 1.0.6 (which does not use the new `jakarta` package). So far, RestAssured 5.2.0 supports the new JAXB specification (with `jakarta.xml.bind` package) but not the new JSON-B speicification (`jakarta.json.bind`).

Any idea when the nez JSON-B specification would be supported?

Thanks

elunji:
Attempt to resolve #1651


johanhaleby:
Unfortunately, I cannot accept this because rest-assured is still compatbile with Java 8 and I see that you've bump the version to 11. We probably should get off Java 8 though, I guess it's time :) But I've struggled with building the project (especially the Groovy stuff) with Java 17+. Does that work for you?

elunji:
Thanks for looking. Regarding java 8 vs Java 11, java 11 is indeed needed to compile with yasson 3.0.3. However, the target classes are still produced with java 8 bytecode. I attempted to take the smallest possible step. Maybe we could use java 11 only for the yasson tests and not the rest of the code, but the would require more changes. Regarding java 17, indeed, it does not build (it builds with Java 11), but this due to the old version of mockito which is not compatible with Java 17. Running tests with Java 17 would require to upgrade mockito (and rewrite some of the tests). I can have a look, but this is almost a different PR.

PR #1861, opened 2026-02-14
Add support for Jackson3ObjectMapperFactory in JsonPath

Anusha-7254:
Added a small comment to Greeting.java example for clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,14 @@ languages into the Java domain.
 
 ## Examples
 Here's an example of how to make a GET request and validate the JSON or XML response:
+### Important
+
+Before using REST Assured methods like get(), given(), then(), etc.,
+make sure to add the required static imports:
+
+import static io.restassured.RestAssured.*;
+import static org.hamcrest.Matchers.*;
+
 
 ```java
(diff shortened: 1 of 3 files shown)


KP-EV:
@johanhaleby could we get any clarification if this is getting merged or not? :) This bug prevents us from fully migrating to jackson3 since rest assured still uses jackson2 for jsonpath under the hood

thanks

johanhaleby:
Thanks for this, Anusha. The `using(Jackson3ObjectMapperFactory)` overload was a real gap. `JsonPath` already had the Gson, Jackson 1 and Jackson 2 overloads but not Jackson 3, even though `JsonPathConfig` supports it, so I wanted that part in.

I've pushed it to master in 2054f372 with you as co-author. I fixed the duplicate import and the indentation while moving it over.

I left the other two changes out. The README section repeated import guidance that's already in the README, and the imports weren't inside a code block, and the comment in `Greeting.java` wasn't needed. Since the overload went straight to master rather than through this PR, I'm closing this one.

Thanks again for catching the missing overload.

PR #1878, opened 2026-07-22
Spread properties over object lists even when the first element lacks it

### LINKED ISSUE #1875: JsonPath: array.properties GPath spread returns null (drops all data) when the first array element lacks a properties key (Groovy 5 regression)
### Description

For a JSON array of objects, `array.properties` should behave like Groovy-4 GPath spread, `array.collect { it['properties'] }`, yielding `null` for elements that lack the key. The Groovy-5 workaround that restores this gates the entire spread on element `[0]`: it runs only when `get(0)` is a `Map` containing `"properties"`. When the first element has no `properties`, `getProperties()` returns `null` and discards every element's data, with no exception.

```
JSON features = [ {geometry:{}}, {properties:{gridId:AB}}, {properties:{gridId:CD}} ]
getList("features.properties")         =>  null   (all data dropped)
getList("features.properties.gridId")  =>  null
```

This is the GeoJSON `FeatureCollection` shape named in the class's own javadoc. GeoJSON allows a feature's `properties` to be absent, and feature order is not guaranteed, so one leading feature without `properties` wipes the whole collection.

### rest-assured version

- Regression introduced in 6.0.0 (json-path Java migration, PR #1844 / commit `1d3db214`, 2025-12-12); reproduced on `master` HEAD `eeed4998` (6.0.2-SNAPSHOT), Groovy 5.0.3, JDK 24. The pre-migration 5.5.6 line used the old GroovyShell path and is not affected.

### Steps to reproduce

```java
import io.restassured.path.json.JsonPath;
import java.util.List;

public class Repro {
    public static void main(String[] a) {
        // first element has NO "properties"
        String bug = "{\"features\":[{\"geometry\":{}},{\"properties\":{\"gridId\":\"AB\"}},{\"properties\":{\"gridId\":\"CD\"}}]}";
        System.out.println("first-missing  features.properties        = " + JsonPath.from(bug).getList("features.properties"));
        System.out.println("first-missing  features.properties.gridId = " + JsonPath.from(bug).getList("features.properties.gridId"));

        // SAME data, only the order differs (properties-less element no longer first)
        String ok = "{\"features\":[{\"properties\":{\"gridId\":\"AB\"}},{\"geometry\":{}},{\"properties\":{\"gridId\":\"CD\"}}]}";
        System.out.println("first-present  features.properties        = " + JsonPath.from(ok).getList("features.properties"));

        // ordinary (non-'properties') key spreads correctly regardless of order
        String foo = "{\"features\":[{\"geometry\":{}},{\"foo\":{\"gridId\":\"AB\"}},{\"foo\":{\"gridId\":\"CD\"}}]}";
        System.out.println("ordinary key   features.foo               = " + JsonPath.from(foo).getList("features.foo"));
    }
}
```

Output:

```
first-missing  features.properties        = null                          <-- BUG
first-missing  features.properties.gridId = null                          <-- BUG
first-present  features.properties        = [{gridId=AB}, null, {gridId=CD}]   <-- same data, order swapped: correct
ordinary key   features.foo               = [null, {gridId=AB}, {gridId=CD}]   <-- ordinary key: correct
```

The cleanest isolation: inputs 1 and 3 have identical key name and data and differ only in array order, yet one returns total `null` and the other the correct `[{gridId=AB}, null, {gridId=CD}]`. Swapping only the leaf key `properties` for `foo` (same shape) already works, so Groovy 5 handles the ordinary spread; the loss is specific to the `properties` workaround's first-element guard.

### Expected

`features.properties` = `[null, {gridId=AB}, {gridId=CD}]` and `features.properties.gridId` = `[AB, CD]`, per the class javadoc contract (`list.collect { it['properties'] }`) and the intent of `JsonPathTest.automatically_escapes_json_attributes_whose_name_equals_properties()`.

### Actual

`null`. The whole spread disappears whenever the first element lacks `properties`.

### Root cause

`json-path/src/main/java/io/restassured/internal/path/json/Groovy5JsonSlurperWorkarounds.java:57`, `ProxyArray#getProperties()`:

```java
public Object getProperties() {
  if (!isEmpty() && get(0) instanceof Map<?, ?> entry0 && entry0.containsKey(PROPERTIES)) {  // <-- guards on element[0]
    return stream().map(elem -> {
      if (elem instanceof Map<?, ?> map && map.containsKey(PROPERTIES)) {
        return map.get(PROPERTIES);
      } else {
        return null;                       // inner body ALREADY tolerates absent/non-map entries
      }
    }).toList();
  }
  return null;                             // <-- first-element guard failed -> WHOLE spread dropped
}
```

The inner lambda already maps absent and non-map entries to `null`, so the spread handles heterogeneous lists on its own. The sole cause is the `get(0) ... containsKey(PROPERTIES)` guard, which samples element `[0]` alone to decide whether to run at all. The existing unit test's fixture has `properties` on the first feature, so no test exercises this heterogeneous path.

### Suggested fix

Gate on the key being present anywhere in the list (or on the list containing any `Map`), instead of element `[0]`:

```java
public Object getProperties() {
  boolean anyMapHasProperties = stream()
      .anyMatch(e -> e instanceof Map<?, ?> m && m.containsKey(PROPERTIES));
  if (anyMapHasProperties) {
    return stream()
        .map(e -> (e instanceof Map<?, ?> m && m.containsKey(PROPERTIES)) ? m.get(PROPERTIES) : null)
        .toList();
  }
  return null;
}
```

Add a regression test with a heterogeneous `features` array whose first element lacks `properties` (plus an order-swapped variant), asserting the full spread.

---
*Found via property-based & differential bug-hunting, part of an effort to scale PBT (DepTyCheck-based) testing across the OSS ecosystem.*

*If this is intended / by-design: I'm really sorry, please just close it — no need to flag or ban me. I'm trying to scale property-based testing across the whole ecosystem and my publishing agents may have gotten this one wrong. I read every issue and follow up on each.*

HDPark95:
## What & why

`features.properties` on a list of objects returns `null` when the first element has no `properties` key. This is a 6.0.0 regression from the Groovy 5 migration workarounds: `Groovy5JsonSlurperWorkarounds.getProperties()` gates the whole spread on element `[0]` containing the key. In GeoJSON FeatureCollections a feature may legally omit `properties` and feature order carries no guarantees, so a leading geometry-only feature wiped the entire spread.

## Fix

Detect the JSON-field spread by checking whether **any** element carries the key (`stream().anyMatch(...)`) instead of only element `[0]`, restoring the Groovy 4 behaviour the workaround exists to preserve.

## Tests

- New `JsonPathTest#spreads_properties_over_list_of_objects_even_when_the_first_element_lacks_the_properties_key`: GeoJSON-style collection whose first feature has no `properties`. Verified it fails on current master (reproduces #1875) and passes with the fix.
- Full `json-path` module test suite green locally.
- `changelog.txt` entry added under "next version".

Fixes #1875

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 Changelog next version
 ----------------------
+* Fix JsonPath "properties" spread over a list of objects returning null when the first element lacks a "properties" key (e.g. a GeoJSON feature carrying only a geometry), a regression from the Groovy 5 migration (fixes #1875)
 * Strip sensitive headers (Authorization and Cookie) when a redirect crosses to a different host, so a token or session isn't leaked to the redirect target. This is on by default and can be turned off with RedirectConfig.stripSensitiveHeadersOnCrossHostRedirect(false) (#1873). Thanks to the University of Sydney security research team (Liyi Zhou, Ziyue Wang, Strick, Maurice, and Chenchen Yu) for the report.
 
 Changelog 6.0.1 (2026-07-10)
--- a/json-path/src/main/java/io/restassured/internal/path/json/Groovy5JsonSlurperWorkarounds.java
+++ b/json-path/src/main/java/io/restassured/internal/path/json/Groovy5JsonSlurperWorkarounds.java
@@ -54,9 +54,13 @@ class Groovy5JsonSlurperWorkarounds {
     private static final long serialVersionUID = 1L;
 
     public Object getProperties() {
-      if (!isEmpty() && get(0) instanceof Map<?, ?> entry0 && entry0.containsKey(PROPERTIES)) {
+      if (!isEmpty() && stream().anyMatch(elem -> elem instanceof Map<?, ?> map && map.containsKey(PROPERTIES))) {
         // Groovy 4–style JSON behaviour:
         // [map, map].properties -> [map['properties'], ...]
+        // Any element carrying a "properties" key marks this as a JSON-field spread rather than a
+        // meta-property access. Gating on element [0] alone dropped the whole spread when the first
+        // element happened to lack "properties" (e.g. a GeoJSON feature with only a geometry), even
+        // though feature order is not guaranteed and later elements did carry it.
         return stream().map(elem -> {
           if (elem instanceof Map<?, ?> map && map.containsKey(PROPERTIES)) {
             return map.get(PROPERTIES);
--- a/json-path/src/test/java/io/restassured/path/json/JsonPathTest.java
+++ b/json-path/src/test/java/io/restassured/path/json/JsonPathTest.java
@@ -828,6 +828,16 @@ public class JsonPathTest {
 
         // Then the spread yields the values of the features that do carry "properties" instead of null
         assertThat(jsonPath.getList("features.properties.gridId", Integer.class), hasItems(6, 7));
+
+        // hasItems above only proves the surviving values are present; it cannot distinguish
+        // [null, 6, 7] from [6, 7]. Assert features.properties directly so the positional-null
+        // contract is pinned: the properties-less first feature must keep its slot as a leading
+        // null rather than being dropped, leaving a 3-element list of [null, {gridId=6}, {gridId=7}].
+        List<Map<String, Object>> propertiesByFeature = jsonPath.getList("features.properties");
+        assertThat(propertiesByFeature, hasSize(3));
+        assertThat(propertiesByFeature.get(0), is(nullValue()));
+        assertThat(propertiesByFeature.get(1), hasEntry("gridId", (Object) 6));
+        assertThat(propertiesByFeature.get(2), hasEntry("gridId", (Object) 7));
     }
 
     @Test public void
--- a/json-path/src/test/java/io/restassured/path/json/JsonPathTest.java
+++ b/json-path/src/test/java/io/restassured/path/json/JsonPathTest.java
@@ -787,6 +787,50 @@ public class JsonPathTest {
     }
 
     @Test public void
+    spreads_properties_over_list_of_objects_even_when_the_first_element_lacks_the_properties_key() {
+        // Given a GeoJSON-style FeatureCollection whose FIRST feature has no "properties".
+        // GeoJSON permits an absent "properties" and does not guarantee feature order, so a leading
+        // feature without "properties" must not wipe the whole spread (issue #1875).
+        String json = """
+                {
+                   "features":[
+                      {
+                         "type":"Feature",
+                         "geometry":{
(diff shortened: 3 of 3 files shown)


renechoi (commented):
Nicely scoped fix, and the #1875 diagnosis matches what's in the code.

What makes this safe to take is that the new guard is strictly weaker than the old one: whenever `get(0) instanceof Map && entry0.containsKey(PROPERTIES)` held, `stream().anyMatch(...)` necessarily holds too, and the inner mapping body is unchanged. So every input that already produced a spread produces the identical list, and the two guards diverge only when some but not the first element carries `properties`, which is exactly the drop-to-`null` case #1875 reports. It can only turn a buggy `null` into the correct spread, not alter output that was already correct. The "no element anywhere has `properties`" path still returns `null`, so the meta-property fallback this workaround exists for stays untouched.

Minor, take it or leave it: with `anyMatch` the leading `!isEmpty()` is now redundant, since `anyMatch` returns `false` on an empty stream anyway. Harmless as a fast-path, so no need to touch it.

One thing worth weighing on the test before it goes in. It asserts `getList("features.properties.gridId", Integer.class)` with `hasItems(6, 7)`, which does prove the values are no longer dropped. But that is a containment check a level below where the bug lives (`features.properties` / `ProxyArray.getProperties()`), so it cannot distinguish `[null, 6, 7]` from `[6, 7]`: if some later change ever filtered the properties-less element instead of mapping it to `null`, the list would silently shrink from 3 to 2 and this test would stay green. Would it be worth also asserting on `features.properties` directly, where I would expect a 3-element list with a leading `null` (`[null, {gridId=6}, {gridId=7}]`)? That pins the positional-`null` contract the issue calls out (`list.collect { it['properties'] }`), which is the actual guarantee here. The first-present direction is already covered by `automatically_escapes_json_attributes_whose_name_equals_properties`, so a separate order-swapped fixture probably adds little.

HDPark95:
Good catch. `hasItems(6, 7)` only proves the two surviving values are present, so it can't tell `[null, 6, 7]` from `[6, 7]`: a later change that filtered the properties-less feature instead of mapping its slot to `null` would silently shrink the list to two and keep the test green.

06ca7f69 adds a direct assertion on `features.properties`, pinning the full positional result to `[null, {gridId=6}, {gridId=7}]` (3 elements, leading `null` at index 0). That locks the positional contract the issue calls out (`list.collect { it['properties'] }`).

On `!isEmpty()`: agreed it's redundant now that the guard is `anyMatch`, but I left it in as a cheap fast-path so the empty-list case skips the stream entirely.

PR #1880, opened 2026-07-22
Fix nested properties navigation over a list of objects in JsonPath (#1879)

### LINKED ISSUE #1879: JsonPath: nested `properties` over a list of objects returns null / leaks Groovy meta-data because the list spread result is not the proxy list (Groovy 5 regression)
### Description

For a JSON list of objects, `list.properties` is treated as a Groovy-4-style GPath spread and returns `list.collect { it['properties'] }`. This is restored on Groovy 5 by `Groovy5JsonSlurperWorkarounds.ProxyArray` (issue #552, PR #1844), and a single-level spread works today. Nested `properties` over a **map** also works and has a test (`parses_immediate_and_nested_properties_in_map_structure_not_wrapped_in_list`, `properties.properties.height`).

The natural combination of the two, a nested `properties` over a **list**, silently breaks. The spread result is a plain immutable `List` (from `Stream.toList()`) rather than the `ProxyArray` proxy, so the Groovy 5 workaround does not propagate to it. A second `properties` hop then resolves to Groovy's meta-property instead of the JSON field:

- `list.properties.properties.x` returns `null`, and
- `list.properties.properties` returns the list's Groovy meta-data, `{class=class java.util.ImmutableCollections$ListN, empty=false}`.

That `{class, empty}` map is exactly the symptom reported (and long since fixed for the single-level case) in issue #385. It resurfaces here for the nested-over-list case.

### rest-assured version

Regression introduced in 6.0.0 (json-path Java migration, PR #1844). Reproduced on `master` HEAD `eeed4998` (`6.0.2-SNAPSHOT`), Groovy `5.0.3`, JDK 17. The pre-migration 5.5.x line used the old GroovyShell path and is not affected. This is independent of #1875: that fix (PR #1878) changes only the spread's element gate, not the type of the spread result, so the nested case stays broken after it.

### Steps to reproduce

```java
import io.restassured.path.json.JsonPath;

public class Repro {
    public static void main(String[] args) {
        // Works: nested "properties" over a MAP (existing test)
        String map = "{\"properties\":{\"properties\":{\"height\":5}}}";
        System.out.println("map   properties.properties.height       = "
                + JsonPath.from(map).getInt("properties.properties.height"));

        // Works: single-level "properties" spread over a LIST
        String single = "{\"items\":[{\"properties\":{\"gridId\":6}},{\"properties\":{\"gridId\":7}}]}";
        System.out.println("list  items.properties.gridId            = "
                + JsonPath.from(single).getList("items.properties.gridId"));

        // BUG: nested "properties" over a LIST
        String nested = "{\"items\":[{\"properties\":{\"properties\":{\"gridId\":6}}},"
                + "{\"properties\":{\"properties\":{\"gridId\":7}}}]}";
        System.out.println("list  items.properties.properties.gridId = "
                + JsonPath.from(nested).get("items.properties.properties.gridId"));
        System.out.println("list  items.properties.properties        = "
                + JsonPath.from(nested).get("items.properties.properties"));
        System.out.println("list  items.properties (one hop)         = "
                + JsonPath.from(nested).get("items.properties"));
    }
}
```

Output:

```
map   properties.properties.height       = 5
list  items.properties.gridId            = [6, 7]
list  items.properties.properties.gridId = null                                          <-- BUG (expected [6, 7])
list  items.properties.properties        = {class=class java.util.ImmutableCollections$ListN, empty=false}   <-- BUG (Groovy meta-data leaked)
list  items.properties (one hop)         = [{properties={gridId=6}}, {properties={gridId=7}}]
```

The map form on line 1 and the list form on the buggy lines navigate the same key at the same depth; only the container (map vs list) differs, yet the list form drops the data and leaks meta-data.

JUnit reproduction:

```java
@Test
public void spreads_nested_properties_over_a_list_of_objects() {
    String json = """
            {
               "items":[
                  { "properties": { "properties": { "gridId": 6 } } },
                  { "properties": { "properties": { "gridId": 7 } } }
               ]
            }""";

    List<Integer> ids = new JsonPath(json).getList("items.properties.properties.gridId", Integer.class);

    assertThat(ids, contains(6, 7)); // fails on master: iterable did not contain <6>
}
```

### Expected behaviour

`items.properties.properties.gridId` yields `[6, 7]`, mirroring the map form `properties.properties.height`. No Groovy meta-data (`class`, `empty`) is ever exposed as a JsonPath result.

### Root cause and suggested fix

`Groovy5JsonSlurperWorkarounds.ProxyArray.getProperties()` builds the spread with `stream().map(...).toList()`, which returns a plain immutable `List`, not a `ProxyArray`. Because the result is not the proxy type, the `properties` workaround stops applying after the first list hop.

Collecting the spread into a `ProxyArray` (the same proxy the parser produces) instead of `toList()` makes the result carry the workaround forward. I confirmed locally that this returns `[6, 7]` for the nested case and keeps the existing `JsonPathTest` suite green.

kdelay:
## What

Fixes #1879.

On Groovy 5, `JsonPath` navigation of the form `items.properties.properties.gridId` (a nested `properties` field over a list of objects) returned `null` or leaked Groovy meta-data such as `{class, empty}` instead of the JSON values.

## Why

`Groovy5JsonSlurperWorkarounds.ProxyArray.getProperties()` spreads a list of objects onto their `"properties"` field, but it collected the result with `stream().map(...).toList()`, producing a plain immutable `List`. A subsequent `.properties` hop on that result therefore hit Groovy meta-properties again (the very thing the workaround exists to block), so the second navigation broke.

The fix collects the spread into a `ProxyArray` so the workaround is carried forward across chained `"properties"` navigations. This is independent of the element-gate change in #1878 (which only decides *which* elements are spread, not the *type* of the spread result) and applies identically to the current and post-#1878 versions of the method.

## Verification

- Added `spreads_nested_properties_over_a_list_of_objects` to `JsonPathTest`. It FAILS on unmodified `master` (`Expected iterable containing [<6>, <7>] but no item was <6>`) and PASSES with the fix.
- `mvn -pl json-path -am verify` (JDK 17): `JsonPathTest` 74/74 green, full json-path module 102/102 green.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 Changelog next version
 ----------------------
+* Fix JsonPath returning null or Groovy meta-data for a nested "properties" navigation over a list of objects (e.g. items.properties.properties.x) on Groovy 5, because the list spread result was a plain list instead of the proxy list that carries the "properties" workaround forward (fixes #1879)
 * Strip sensitive headers (Authorization and Cookie) when a redirect crosses to a different host, so a token or session isn't leaked to the redirect target. This is on by default and can be turned off with RedirectConfig.stripSensitiveHeadersOnCrossHostRedirect(false) (#1873). Thanks to the University of Sydney security research team (Liyi Zhou, Ziyue Wang, Strick, Maurice, and Chenchen Yu) for the report.
 
 Changelog 6.0.1 (2026-07-10)
--- a/json-path/src/main/java/io/restassured/internal/path/json/Groovy5JsonSlurperWorkarounds.java
+++ b/json-path/src/main/java/io/restassured/internal/path/json/Groovy5JsonSlurperWorkarounds.java
@@ -57,13 +57,17 @@ class Groovy5JsonSlurperWorkarounds {
       if (!isEmpty() && get(0) instanceof Map<?, ?> entry0 && entry0.containsKey(PROPERTIES)) {
         // Groovy 4–style JSON behaviour:
         // [map, map].properties -> [map['properties'], ...]
-        return stream().map(elem -> {
+        // Collect into a ProxyArray (not a plain List) so that a further "properties" hop on the
+        // spread result keeps navigating the JSON field instead of leaking Groovy meta-data.
+        ProxyArray spread = new ProxyArray();
+        for (Object elem : this) {
           if (elem instanceof Map<?, ?> map && map.containsKey(PROPERTIES)) {
-            return map.get(PROPERTIES);
+            spread.add(map.get(PROPERTIES));
           } else {
-            return null;
+            spread.add(null);
           }
-        }).toList();
+        }
+        return spread;
       }
       return null;
     }
--- a/json-path/src/test/java/io/restassured/path/json/JsonPathTest.java
+++ b/json-path/src/test/java/io/restassured/path/json/JsonPathTest.java
@@ -787,6 +787,22 @@ public class JsonPathTest {
     }
 
     @Test public void
+    spreads_nested_properties_over_a_list_of_objects() {
+        // items.properties.x (single spread) and properties.properties.x (nested map) both work;
+        // the nested list form must too. The list "properties" spread result must stay a proxy list
+        // so the second "properties" hop navigates the JSON field, not Groovy meta-data (issue #1879).
+        String json = """
+                {
+                   "items":[
+                      { "properties": { "properties": { "gridId": 6 } } },
+                      { "properties": { "properties": { "gridId": 7 } } }
+                   ]
+                }""";
+        JsonPath jsonPath = new JsonPath(json);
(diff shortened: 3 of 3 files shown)


renechoi (commented):
Good analysis of the chaining behavior in Groovy 5 workaround.

Collecting the spread into a `ProxyArray` instead of a plain `List` ensures that subsequent `.properties` access on the result routes back into `ProxyArray.getProperties()` rather than reverting to default Groovy meta-property reflection. Because `ProxyArray` extends `ArrayList`, returning `spread` maintains full compatibility with standard `List` consumers while preserving the workaround for arbitrarily nested `properties` hops (`items.properties.properties...`).

Also verified that this change cleanly composes with #1878: #1878 changes the outer predicate (`anyMatch`), while this PR changes the return type inside the block.

PR #1881, opened 2026-07-24
Fix PathSupport URL parsing when query parameter contains another URL (#1836)

### LINKED ISSUE #1836: Bug: URLs with URLs as query parameters lead to `java.net.MalformedUrlException`
Description:
If an URL is called with the another URL in a query parameter, that does not have a `/` before the query param a `MalformedURLException` is thrown.

So for example:

1. `https://example.com/?redirect=https://example.com/callback` works ✅ 
2. `https://example.com?redirect=https://example.com/callback` fails ❌ 

I think this is a bug in the way this method is implemented: https://github.com/rest-assured/rest-assured/blob/d3fa59c145097811d18c41a03ff27240cb75edfb/rest-assured/src/main/groovy/io/restassured/internal/support/PathSupport.groovy#L42-L54

Because after the initial `https://` is removed somewhere before, it will still think the URL is fully qualified even though it should not.

Reproduction code:
Kotlin:
```kotlin
    @Test
    fun reproduceUrlBug() {
        When {
            get("https://example.com?redirect=https://example.com/callback")
        }
    }
```
Java:
```java
    @Test
    public void reproduceUrlBug() {
        when().get("https://example.com?redirect=https://example.com/callback");
    }
```

Exception:
```
java.net.MalformedURLException: no protocol: example.com?redirect=https://example.com/callback
```

renechoi:
## What

Fixes #1836.

Fixes MalformedURLException thrown when a target URL contains another URL in a query parameter without a trailing slash before the query string (e.g. `https://example.com?redirect=https://example.com/callback`).

## Why

In `PathSupport.groovy`, `isFullyQualified(targetUri)` and `getPath(targetUri)` searched for the first path slash `/` after scheme in `targetUri` rather than in `path` (the URL string before `?`).

When `targetUri` had no path slash before `?` (such as `https://example.com?redirect=...`), line 66 of `getPath()` (`StringUtils.indexOf(targetUri, "/", indexOfScheme + 3)`) matched the slash inside the query parameter (`?redirect=https://...`), returning `/example.com/callback` instead of `/`. Similarly, `isFullyQualified("example.com?redirect=https://...")` incorrectly returned `true`.

By scoping the slash search to `path` (the substring before `?`), `getPath()` returns `/` and `isFullyQualified()` accurately determines if `path` itself has a scheme and path.

## Verification

- Added `returns_slash_when_fully_qualified_uri_has_no_path_and_query_param_has_url`, `correctly_identifies_fully_qualified_uri_when_no_path_and_query_param_has_url`, and `correctly_identifies_non_fully_qualified_uri_when_query_param_has_url` to `PathSupportTest`.
- `mvn -pl rest-assured -am test`: all 188 tests in module `rest-assured` pass.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 Changelog next version
 ----------------------
+* Fix PathSupport throwing MalformedURLException when target URL has no path slash before a query parameter containing another URL (fixes #1836)
 * Strip sensitive headers (Authorization and Cookie) when a redirect crosses to a different host, so a token or session isn't leaked to the redirect target. This is on by default and can be turned off with RedirectConfig.stripSensitiveHeadersOnCrossHostRedirect(false) (#1873). Thanks to the University of Sydney security research team (Liyi Zhou, Ziyue Wang, Strick, Maurice, and Chenchen Yu) for the report.
 
 Changelog 6.0.1 (2026-07-10)
--- a/rest-assured/src/main/groovy/io/restassured/internal/support/PathSupport.groovy
+++ b/rest-assured/src/main/groovy/io/restassured/internal/support/PathSupport.groovy
@@ -44,8 +44,9 @@ class PathSupport {
       return false
     }
 
-    def indexOfFirstSlash = targetUri.indexOf("/");
-    def indexOfScheme = targetUri.indexOf("://");
+    def path = StringUtils.substringBefore(targetUri, "?")
+    def indexOfFirstSlash = path.indexOf("/")
+    def indexOfScheme = path.indexOf("://")
     if (indexOfScheme == -1) {
       // If we didn't find a single :// in the path then we know that the targetUri is not fully-qualified
       return false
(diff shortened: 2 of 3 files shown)


kdelay (commented):
The root cause analysis looks right. Both isFullyQualified and getPath were scanning the raw targetUri, so a query parameter that itself contains :// and a / shifted indexOfScheme, indexOfFirstSlash, and indexOfPath into the query string. Scoping everything to the pre-query path is a clean fix and keeps the two methods consistent.

One test-coverage gap worth closing: the three new tests all exercise the no-path case (host directly followed by ?). The regression only surfaced when there was no real path before the query, but a positive assertion would pin the behavior down, e.g.

    getPath("https://example.com/real/path?redirect=https://other.com/cb")

should return "/real/path". That guards against the embedded-URL slash in the query being mistaken for the target path even when a genuine path is present, which is exactly the confusion the old targetUri-based indexOf produced.

renechoi:
Added in 2595ad3. Both cases are in now: `getPath("https://example.com/real/path?redirect=https://other.com/cb")` asserting `/real/path`, plus the `isFullyQualified` counterpart for the same URI.

Worth recording what they actually catch, since I measured it instead of assuming. Reverting `PathSupport.groovy` to master while keeping all five tests fails exactly two, and both are from the original no-path set:

- `returns_slash_when_fully_qualified_uri_has_no_path_and_query_param_has_url` returns `//example.com/callback` instead of `/`
- `correctly_identifies_non_fully_qualified_uri_when_query_param_has_url` returns `true` instead of `false`

The two new with-path cases pass on the old code as well. On master `getPath` resolved `indexOfScheme` against the pre-query substring but `indexOfPath` against the raw target URI, and when a genuine path is present the first slash after the scheme sits inside that path, so the raw scan landed on the same index anyway. The fault only surfaced when the host was followed directly by `?`.

So they are not regression catchers for #1836. They guard the refactor: the fix moves `indexOfPath` onto the pre-query substring, and these pin the with-path branch that change touches. I am happy to drop them if you would rather keep the test set strictly to the cases that fail without the fix.

`PathSupportTest` is 15/15 and the `rest-assured` module suite is 190/190 on JDK 21.

